PR #20, opened 2020-03-31
wwtlib/Planets.cs: see if we can use JPGs instead of PNGs for the planet textures

pkgw:
JPEG files are much smaller, but don't support transparency.


astrojonathan:
Besides having no Alpha , PEG's are also lossy.

pkgw:
Unfortunately there's no magic here and the JPEGs do render with black squares around them. That's a bit unfortunate since the PNG files are fairly large and are loaded upon startup of the webclient, increasing the size of the initial data transfer. Possible solutions might be to lazy-load or to somehow downscale when appropriate, since in many cases these textures are just being rendered a handful of pixels wide.

PR #22, opened 2020-04-01
azure-pipelines.yml: also publish version tags to a `major.minor` directory

pkgw:
When a pull request is merged to master, it will be uploaded to the `engine/latest` directory on the Azure Storage. When a version tag is pushed, it will currently be uploaded to (e.g.) `engine/1.2.3`. Here we add another stanza to also uploaded tagged release to a "major.minor" directory, e.g. `engine/1.2`. This way, a web app that wants to get bugfixes can just target the `1.2` directory and not have to track every individual release. This is of course assuming a semver model where micro updates do not break anything.

We may need to purge the CDN cache upon releases in order for this workflow to fully work. It looks like that can be automated as well.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -18,9 +18,15 @@ jobs:
       if($branch.StartsWith("refs/tags/v")) {
         $version = $branch.Substring(11)
         Write-Host "##vso[task.setvariable variable=pub_prefix;]engine/$version"
+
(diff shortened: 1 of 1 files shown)


pkgw:
Note: CI logs show the right behavior ­— the second CI step is skipped here (thought it would also have been skipped because pull requests don't upload anyway)

PR #42, opened 2020-05-23
Monorepo-ification with new embed tool

pkgw:
After much experimentation, I think I have a workflow that will enable a reasonable release and CI/CD workflow in a monorepo scheme.

This PR adds the new WWT embed application and the associated embed creator application. In the process of prototyping these, it became pretty clear that a monorepo approach was going to result in a massively more straightforward development workflow if the issues surrounding releases and versioning could be worked out. I think that I've devised something that works to help with that.


pkgw:
I think we're at the point where we have to start iterating the CI/CD scripts on the live repo. Here we go ...

PR #52, opened 2020-06-17
fix(engine): don't attempt to proxy localhost URLs on nonstandard ports

pkgw:
We preloaded the engine URL helpers to know not to try to proxy the localhost domain, but it turns out that `localhost:8080` is treated separately from just `localhost`. Add some logic to detect these domains and recognize them as localhost-y if needed.

Without this, if you try to load files off of something like `localhost:8080`, e.g. a local tiled image for testing, 404's can cause the engine to start trying to proxy everything which breaks all further downloads.


pkgw:
Confirmed in local testing that this fixes my bug.

PR #68, opened 2020-10-18
HiPS port from Windows client

imbasimba:
Ongoing work. Just got it to compile successfully, however many remaining issues until finished.


pkgw (commented):
This looks great!

One small-bore comment: the changes to `RenderContext.cs` and `wwtlib.csproj` show up as rewriting the entire files, because the line endings inside Git are being changed from Unix-y to Windows-y. Would you be comfortable rewriting the history to eliminate the line-ending changes? I know that there are `.gitattributes` settings that can control how Git stores things internally vs. how they're expressed on the filesystem — but since I don't use Windows I'm afraid that I'm not familiar with the optimal settings.

imbasimba:
I have updated all files to use LF now. It turns out there is also a handy visual studio switch for changing the end of line character

PR #71, opened 2020-12-06
Solved gap issue for most use cases

imbasimba:
-Minified gap for order 0-2 & eliminated gap for order 2-7.  Higher orders gaps are already practically invisible due to their tiny size.
-Minor fixes
Fixed texture skew, problem with IsTileBigEnough, usage of Tile cache, dealing with 404 tiles


pkgw:
I think we have a winner! Playing around with HiPS datasets in the webclient, the experience is smooth and artifact-free, except for a few datasets that are in Galactic coordinates that are rendered in equatorial.

Congratulations, this is great work!

PR #72, opened 2020-12-10
Galactic HiPS & Basic HiPS properties support

imbasimba:


engine/wwtlib/HipsProperties.cs
```diff
@@ -0,0 +1,213 @@
...
+namespace wwtlib
+{
+
+    public class HipsProperties
+    {
+        public Dictionary<string, string> Properties = new Dictionary<string, string>();
+        public VoTable VoTable = null;
+        public bool downloadComplete = false;
+        private WebFile webFile;
+        public static HipsProperties GetPropertiesCat(string url)
```
pkgw, line 13: Is this function used at all? I don't think so.

engine/wwtlib/HipsProperties.cs
```diff
@@ -0,0 +1,213 @@
...
+            webFile = new WebFile(propsUrl);
+            webFile.OnStateChange = LoadData;
+            webFile.Send();
+        }
+
+        private void LoadData()
+        {
+            if (webFile.State == StateType.Error)
+            {
+                Script.Literal("alert({0})", webFile.Message);
```
pkgw, line 116: We shouldn't be alert()ing.

engine/wwtlib/HipsProperties.cs
```diff
@@ -0,0 +1,213 @@
...
+            }
+            if (!props.Properties.ContainsKey("dataproduct_type"))
+            {
+                return false;
+            }
+            if (props.Properties["dataproduct_type"] == "cube")
+            {
+                return false;
+            }
+            if (props.Properties.ContainsKey("client_category") && props.Properties["client_category"].ToLowerCase().IndexOf("solar") > -1)
```
pkgw, line 174: Do you know what this test is all about? It seems odd to me.

pkgw (commented):
Mostly looking good, but I am concerned about the `alert()` in the properties code.

engine/wwtlib/HipsProperties.cs
imbasimba: I thought I was going to need it for Catalog HiPS, but that turned out to be false. I'll remove it

engine/wwtlib/HipsProperties.cs
imbasimba: Agreed

engine/wwtlib/HipsProperties.cs
imbasimba: Hmm... Sure looks unnecessary to me. The function itself is not yet used, but I'll have to take a closer look when it is time to fully activate this part.

PR #78, opened 2021-01-12
Integrate the research app and start working on features

pkgw:
Initially I was just going to add more APIs, but the coupling between the research app and the engine is tight enough that we're gonna really regret it if we don't have them all in the same repo. So, let's do it.


pkgw:
Going to merge here to check out the CI wiring-up.

PR #85, opened 2021-03-18
Work on exposing new catalog HiPS APIs to TypeScript

pkgw:
Woohoo catalog HiPS!

CC @imbasimba, FYI these are the kinds of TypeScript changes that are needed to expose new C# APIs.


imbasimba:
Got it :)

PR #86, opened 2021-03-20
Simple FITS HiPS support using JavaScript parsing

imbasimba:
As expected it is quite slow, but it is technically working. 

There seem to be a bug in FitsImage.cs when dealing with the "no data" pixels of the FITS files. For some tiles they are grey, and other they are black. I have not investigated the issue as this entire PR is just a proof of concept for the real solution of reading the data from the shader.


pkgw:
Well, that's quite straightforward. Nice!

PR #88, opened 2021-04-02
FITS HiPS in WebGL

imbasimba:
Basic support for WebGL rendering of FITS HiPS

1. Confirmed to work for 32 & 64 bit float data as well as 16 bit integer data
2. In theory it should also work for 8 bit and 32 bit integer data, but not verified
3. More work needed to support 64 bit integer data
4. WebGL updated to version 2 if supported, otherwise it uses version 1 with JavaScript parsing of FITS data
5. Cuts can be changed through API


pkgw:
OK, time to merge this! I have a few thoughts about things to potentially keep in mind for later, but I don't want to delay this PR any longer since the vast majority looks great.

In particular:

- The `FitsImageWebGL` class feels somewhat redundant with the plain `FitsImage` class, and it would be nice if any duplicated code could be streamlined. But this can be hard with ScriptSharp, I know.
- Are we going to need to add additional ScriptInterface or WWTControl methods that provide ImageSet-based access as opposed to name-based access, as we just needed to do to improve the webclient UX?
- Fleshing out UX for cases where WebGL 2 isn't available.

PR #91, opened 2021-04-26
OpenSpace patch

imbasimba:
1. Option to recursively download Folders in a WTML
2. Added optional parameter to set roll of camera
3. Can now add regular image sets to LayerManager via postMessage / research-app messages
4. Disallowed duplicate image sets (pointing to the same URL) & started using image set url as a semi user friendly identifier

@pkgw Do you know what the TODO comment in Folder.cs:371 is referring to? I removed the "true" thing mentioned in the TODO, so it may cause some problem.

### DIFF
--- a/engine-helpers/src/index.ts
+++ b/engine-helpers/src/index.ts
@@ -285,9 +285,11 @@ export interface GotoTargetOptions {
   trackObject: boolean;
 }
 
-/** Options for [[WWTInstance.loadFitsLayer]]. */
-export interface LoadFitsLayerOptions {
-  /** The URL of the FITS file. */
+/** Options for [[WWTInstance.addImageSetLayer]]. */
+export interface AddImageSetLayerOptions {
+  /** The URL of the FITS file 
+   * OR The URL of the desired image set. This should mach an image set url
+   * previously loaded with [[LoadImageCollection]]. */
   url: string;
 
   /** A name to use for the new layer. */
@@ -546,29 +548,29 @@ export class WWTInstance {
   private collectionLoadedPromises: SavedPromise<string, Folder>[] = [];
   private collectionRequests: Map<string, Folder | null> = new Map();
 
-   /** Load a WTML collection and the imagesets that it contains.
-   *
-   * This function triggers a download of the specified URL, which should return
-   * an XML document in the [WTML collection][wtml] format. Any `ImageSet`
-   * entries in the collection, or `Place` entries containing image sets, will
-   * be added to the WWT instance’s list of available imagery. Subsequent calls
-   * to functions like [[setForegroundImageByName]] will be able to locate the
-   * new imagesets and display them to the user.
-   *
-   * Each unique URL is only requested once. Once a given URL has been
-   * successfully loaded, the promise returned by additional calls will resolve
-   * immediately. URL uniqueness is tested with simple string equality, so if
-   * you really want to load the same URL more than once you could add a
-   * fragment specifier.
-   *
-   * If the URL is not accessible due to CORS restrictions, the request will
-   * automatically be routed through the WWT’s CORS proxying service.
-   *
-   * [wtml]: https://docs.worldwidetelescope.org/data-guide/1/data-file-formats/collections/
-   *
-   * @param url: The URL of the WTML collection file to load.
-   * @returns: A promise that resolves to an initialized Folder object.
-   */
+  /** Load a WTML collection and the imagesets that it contains.
+  *
+  * This function triggers a download of the specified URL, which should return
+  * an XML document in the [WTML collection][wtml] format. Any `ImageSet`
+  * entries in the collection, or `Place` entries containing image sets, will
+  * be added to the WWT instance’s list of available imagery. Subsequent calls
+  * to functions like [[setForegroundImageByName]] will be able to locate the
+  * new imagesets and display them to the user.
+  *
+  * Each unique URL is only requested once. Once a given URL has been
+  * successfully loaded, the promise returned by additional calls will resolve
+  * immediately. URL uniqueness is tested with simple string equality, so if
+  * you really want to load the same URL more than once you could add a
+  * fragment specifier.
+  *
+  * If the URL is not accessible due to CORS restrictions, the request will
+  * automatically be routed through the WWT’s CORS proxying service.
+  *
+  * [wtml]: https://docs.worldwidetelescope.org/data-guide/1/data-file-formats/collections/
+  *
+  * @param url: The URL of the WTML collection file to load.
+  * @returns: A promise that resolves to an initialized Folder object.
+  */
   async loadImageCollection(url: string): Promise<Folder> {
     const curState = this.collectionRequests.get(url);
 
@@ -621,14 +623,16 @@ export class WWTInstance {
 
   // Layers
 
-  /** Load a remote FITS file into a data layer and display it.
+  /** Load an image set or a remote FITS file into a data layer and display it.
    *
    * The FITS file must be downloaded and processed, so this API is
    * asynchronous, and is not appropriate for files that might be large.
+   * 
+   * The image set must have previously been created with [[loadImageCollection]]
    */
-  async loadFitsLayer(options: LoadFitsLayerOptions): Promise<ImageSetLayer> {
+  async addImageSetLayer(options: AddImageSetLayerOptions): Promise<ImageSetLayer> {
     return new Promise((resolve, _reject) => {
-      this.si.loadFitsLayer(options.url, options.name, options.gotoTarget, (layer) => {
+      this.si.addImageSetLayer(options.url, options.name, options.gotoTarget, (layer) => {
         resolve(layer);
       })
     });
--- a/engine-helpers/src/index.ts
+++ b/engine-helpers/src/index.ts
@@ -285,19 +285,38 @@ export interface GotoTargetOptions {
   trackObject: boolean;
 }
 
+
+/** Deprecated, use AddImageSetLayerOptions instead.
+ *  Options for [[WWTInstance.addImageSetLayer]]. */
+export interface LoadFitsLayerOptions {
+  /** The URL of the FITS file. */
+  url: string;
+  
+  /** A name to use for the new layer. */
+  name: string;
+
+  /** Whether to seek the view to the positon of the FITS file on the sky,
+   * if/when it successfully loads. */
+  gotoTarget: boolean;
+}
+
 /** Options for [[WWTInstance.addImageSetLayer]]. */
 export interface AddImageSetLayerOptions {
   /** The URL of the FITS file 
    * OR The URL of the desired image set. This should mach an image set url
    * previously loaded with [[LoadImageCollection]]. */
   url: string;
+  
+  /** Tell WWT what type of layer you are Adding.
+   * OR let WWT try to autodetect the type of the data.
+   * Default, autodetect. */
+  mode: "autodetect" | "fits" | "preloaded";
 
   /** A name to use for the new layer. */
   name: string;
 
   /** Whether to seek the view to the positon of the FITS file on the sky,
-   * if/when it successfully loads.
-   */
+   * if/when it successfully loads. */
   gotoTarget: boolean;
 }
 
--- a/engine-helpers/src/index.ts
+++ b/engine-helpers/src/index.ts
@@ -569,9 +569,11 @@ export class WWTInstance {
   * [wtml]: https://docs.worldwidetelescope.org/data-guide/1/data-file-formats/collections/
   *
   * @param url: The URL of the WTML collection file to load.
+  * @param loadChildFolders When true, this method will recursively
+  * download and unpack the content of all [[Folder]]s contained in the WTML file.
   * @returns: A promise that resolves to an initialized Folder object.
   */
-  async loadImageCollection(url: string): Promise<Folder> {
+  async loadImageCollection(url: string, loadChildFolders?: boolean): Promise<Folder> {
     const curState = this.collectionRequests.get(url);
 
     // If we've already loaded the folder, insta-resolve to it.
@@ -589,7 +591,11 @@ export class WWTInstance {
       // the function.
       const holder: { f: Folder | null } = { f: null };
 
-      holder.f = Wtml.getWtmlFile(url, () => {
+      if (loadChildFolders === undefined) {
+        loadChildFolders = false;
+      }
+
+      holder.f = Wtml.getWtmlFile(url, loadChildFolders, () => {
         // The folder at this URL is now fully loaded.
         const f = holder.f as Folder;
         this.collectionRequests.set(url, f);
--- a/engine-helpers/src/index.ts
+++ b/engine-helpers/src/index.ts
@@ -595,7 +595,7 @@ export class WWTInstance {
         loadChildFolders = false;
       }
 
-      holder.f = Wtml.getWtmlFile(url, loadChildFolders, () => {
+      holder.f = Wtml.getWtmlFile(url, () => {
         // The folder at this URL is now fully loaded.
         const f = holder.f as Folder;
         this.collectionRequests.set(url, f);
@@ -609,7 +609,7 @@ export class WWTInstance {
           // Don't filter out promises for other URLs.
           return true;
         });
-      });
+      }, loadChildFolders);
     }
 
     return new Promise((resolve, reject) => {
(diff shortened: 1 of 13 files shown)


pkgw:
OK, I have started taking a close look at this. I separated out the "roll" API, which is pretty straightforward and independent of the other changes, into #94. I'll update this branch to remove the conflict and also bring this back on to `master`.

Can you please remind me of the motivation for adding the recursive WTML download feature? I worry about people accidentally downloading those 20 megabyte Mars location WTMLs when they don't need to.

imbasimba:
Regarding the recursive download, while I also see the danger of the large Mars WTMLs, I would argue that the current behavior of not loading recursively is quite counter intuitive. I know that at least the Open space people and I have on separate occations fallen into the trap of assuming that 
`loadImageCollection('http://www.worldwidetelescope.org/wwtweb/catalog.aspx?W=vampfeeds')`
would actually add the contained image collections. At least this parameter would implicitely explain why the image sets are not added. So I guess the main reason would be to save more people from making the same mistakes, as well as provide some convenience. With some kind of warning I think we should be fine.

On a side note (if I don't misremember things), even [this file](http://www.worldwidetelescope.org/wwtweb/catalog.aspx?W=hips) with nested folders was not loading the actual image sets. Of course, this use case is way easier to defend than the one were new http requests are initiated automatically.

engine/wwtlib/ScriptInterface.cs
```diff
@@ -320,65 +320,93 @@ public ImageSetLayer LoadFits(string url)
...
                 }
                 if (loaded != null)
                 {
                     loaded(imagesetLayer);
                 }
-            });
 
-            return imagesetLayer;
+                return imagesetLayer;
+            } else
```
pkgw, line 348: Connecting to what I wrote above, I don't feel like we want to have behavior here where everything that isn't a known imageset is treated as a FITS file. I think it will be very useful to have a generic "turn URL into an imageset layer" API but it needs to be future-proof for new data types.

engine/wwtlib/ScriptInterface.cs
```diff
@@ -412,14 +440,12 @@ public void StopTour()
             }
         }
 
-        private Folder imageFolder;
         private string imageUrl;
-        public void LoadImageCollection(string url)
+        public void LoadImageCollection(string url, bool loadChildFolders)
```
pkgw, line 444: I think the new argument needs to be optional to avoid API breakage.

engine/wwtlib/Utilities/Histogram.cs
```diff
@@ -71,6 +71,7 @@ public void Show(Vector2d position)
         public static void UpdateImage(ImageSetLayer isl, double z)
         {
             FitsImage image = isl.ImageSet.WcsImage as FitsImage;
+            if(image == null) { return; }
```
pkgw, line 74: Formatting nit here and below, I think it's best to use newlines for such statements to make them more perceptible on a quick skim:

```c#
if (image == null) {
    return;
}
```

engine/wwtlib/WWTControl.cs
```diff
@@ -28,12 +28,28 @@ static WWTControl()
...
+        private static List<Imageset> ImageSets = new List<Imageset>();
         public static Folder ExploreRoot = new Folder();
         public static string ImageSetName = "";
 
         public IUiController uiController = null;
 
+        public static void AddImageSet(Imageset imagesetToAdd)
+        {
+            foreach(Imageset imageset in ImageSets){
+                if(imageset.ImageSetID == imagesetToAdd.ImageSetID)
```
pkgw, line 40: What's the rationale for adding this rejection of duplicates? And, should it be using the URL instead of the ImageSetId? As far as I can tell the ImageSetID is based on a weak hash of the URL, so I'm not sure if it adds any value.

engine/wwtlib/WWTControl.cs
```diff
@@ -2549,6 +2563,18 @@ public Imageset GetImagesetByName(string name)
             return null;
         }
 
+        public Imageset GetImageSetByUrl(string url)
+        {
+            foreach (Imageset imageset in ImageSets)
+            {
+                if (imageset.Url.ToLowerCase() == url.ToLowerCase())
```
pkgw, line 2570: Was there a specific reason to case-fold the URLs? I feel like it would be better to not do that unless there's a strong reason.

research-app-messages/src/classic_pywwt.ts
```diff
@@ -241,13 +239,17 @@ export interface LoadImageCollectionMessage {
 
   /** The URL of the collection to load. */
   url: string;
+
+  /** Recursively load any child folders. */
+  loadChildFolders: boolean;
```
pkgw, line 244: This should be optional, I think?

research-app/src/App.vue
```diff
@@ -106,17 +106,16 @@ class FitsLayerMessageHandler {
...
 
   constructor(owner: App) {
     this.owner = owner;
   }
 
-  handleCreateMessage(msg: classicPywwt.CreateFitsLayerMessage) {
-    if (this.created)
-      return;
+  handleCreateMessage(msg: classicPywwt.CreateImageSetLayerMessage) {
+    if (this.created) {return;}
```
pkgw, line 116: Here too I'd prefer to keep the `return` on its own line.

pkgw (commented):
Some comments, including an idea about how to approach the general API design.

engine-helpers/src/index.ts
```diff
@@ -285,9 +285,11 @@ export interface GotoTargetOptions {
   trackObject: boolean;
 }
 
-/** Options for [[WWTInstance.loadFitsLayer]]. */
-export interface LoadFitsLayerOptions {
-  /** The URL of the FITS file. */
+/** Options for [[WWTInstance.addImageSetLayer]]. */
```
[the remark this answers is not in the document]
imbasimba: Agreed, I'll fix this and the other API breaking issues in a few days

engine-helpers/src/index.ts
```diff
@@ -285,9 +285,11 @@ export interface GotoTargetOptions {
...
+export interface AddImageSetLayerOptions {
+  /** The URL of the FITS file 
+   * OR The URL of the desired image set. This should mach an image set url
+   * previously loaded with [[LoadImageCollection]]. */
```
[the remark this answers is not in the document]
imbasimba: Indeed, something like this would be good to future-proof the API

engine/wwtlib/Folder.cs
```diff
@@ -275,7 +287,7 @@ public void Refresh()
                 proxyFolder.Parent = Parent;
             }
 
-            proxyFolder.LoadFromUrl(urlField, childReadyCallback);
+            proxyFolder.LoadFromUrlWithErrorCallback(urlField, childReadyCallback, childReadyCallback);
```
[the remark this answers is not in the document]
imbasimba: Yeah, you are reading into it correctly. It is just for graceful handling of potential problems during the download

engine/wwtlib/Layers/LayerManager.cs
```diff
@@ -433,12 +433,21 @@ public static ImageSetLayer AddFitsImageSetLayer(ImageSetLayer layer, string tit
...
+
+        public static string GetNextImageSetName()
+        {
+            return getNextName("Image Set");
+        }
+
+        private static string getNextName(string type){
             int currentNumber = 0;
             foreach (Layer layer in AllMaps["Sky"].Layers)
             {
```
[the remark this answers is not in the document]
imbasimba: I'll look into if that's the case

engine/wwtlib/WWTControl.cs
imbasimba: I only see problems with having duplicates. Especially when there are settings like FITS cuts attached to an image set. 
For example, changing the duplicated imagesets cuts would not reflect in the imageset that is actually showing. 

Yeah, using the URL instad is fine

engine/wwtlib/WWTControl.cs
imbasimba: The only reason I did it was to be more forgiving to users typos

research-app-messages/src/classic_pywwt.ts
imbasimba: Yup

engine-helpers/src/index.ts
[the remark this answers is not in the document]
pkgw: Thanks, I appreciate it. I *think* that it wouldn't be hard to turn the `LoadFitsLayer` into a very thin shim, but if it turns out to add much complexity, we can reconsider.

engine-helpers/src/index.ts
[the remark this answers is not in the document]
pkgw: Cool. I *think* that my proposal above should work pretty well, but if you come up with any improvements, it's definitely not set in stone!

engine/wwtlib/WWTControl.cs
pkgw: OK. I tend to think that we should enforce URL-based uniqueness. I do worry about possible cases where the same image-data URL is associated with different overall Imageset objects. Two examples I can think of:

- There might be uses cases to use the a kind of generic "always give me a black tile" URL for multiple different imagesets
- You might want to imagesets with the same "study" image data to compare different astrometric solutions

But, I do see a lot of value in being able to uniquely identify imagesets by URL, these two examples are definitely corner cases, and you can always uniquify a URL by adding an unused query parameter or fragment.

imbasimba: Hmm, there is also the case of HiPS, with multiple image formats. Depending on how we deal with it there could be multiple imagesets with the same URL and different extensions (FITS, jpg, png)

pkgw: Ah, I see what you're saying. I think that it would be not-unreasonable to use a URL fragment to disambiguate if needed? E.g. `http://.../my/hips#fits` vs. `http://.../my/hips#jpg`.

engine-helpers/src/index.ts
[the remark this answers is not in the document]
imbasimba: Please have another look at how I handled the new LoadFitsLayer shims. I have not yet fully mastered how all these typescript components interact, so there may be something sub optimal lurking there

engine/wwtlib/WWTControl.cs
imbasimba: Yes, something like that could be implemented for HiPS.

At least the ImageSetID has to be unique, since that is currently used to find and access tiles in the TileCache. Imagesets with the same URL will now refer to the exact same Tiles in the draw cycle.
Of course that could be solved by using a composite of traits to create the ID or some other way to create a ID for each imageset.

pkgw:
Relevant to some of the discussion here, @Carifio24 pointed out an example of HiPS imagesets with identical URLs:

```
      <ImageSet DemUrl="" MSRCommunityId="0" MSRComponentId="0" Permission="0" Generic="False" DataSetType="Sky" BandPass="IR" Url="http://surveys.roe.ac.uk/hips71/UHS/UHSDR1_PIX/Norder{0}/Dir{1}/Npix{2}" TileLevels="10" WidthFactor="1" Sparse="False" Rotation="0" QuadTreeMap="0123" Projection="Healpix" Name="UHS DR1 J" FileType="png fits" CenterY="0" CenterX="0" BottomsUp="False" StockSet="False" ElevationModel="False" OffsetX="0" OffsetY="0" BaseTileLevel="0" BaseDegreesPerTile="180" ReferenceFrame="Sky" MeanRadius="1">
        <Credits />

    <ImageSet DemUrl="" MSRCommunityId="0" MSRComponentId="0" Permission="0" Generic="False" DataSetType="Sky" BandPass="IR" Url="http://surveys.roe.ac.uk/hips71/UHS/UHSDR1_CAT/Norder{0}/Dir{1}/Npix{2}" TileLevels="11" WidthFactor="1" Sparse="False" Rotation="0" QuadTreeMap="0123" Projection="Healpix" Name="UHS DR1 J" FileType="tsv" CenterY="0" CenterX="0" BottomsUp="False" StockSet="False" ElevationModel="False" OffsetX="0" OffsetY="0" BaseTileLevel="0" BaseDegreesPerTile="180" ReferenceFrame="Sky" MeanRadius="1">
```

Same base URL, but there seems to be one additional level of tiling available in the tsv format.

engine/wwtlib/Folder.cs
```diff
@@ -275,7 +287,10 @@ public void Refresh()
                 proxyFolder.Parent = Parent;
             }
 
-            proxyFolder.LoadFromUrl(urlField, childReadyCallback);
+            //Also listening to errors, to make sure clients do not wait forever in the case of a 404 or similar.
+            //Especially useful for recursive downloads, where potentially dussins of URL's are downloaded.
```
pkgw, line 291: Thanks, I think this will be useful to discuss in the source! English spelling: "dozens" not "dussins".

research-app-messages/src/classic_pywwt.ts
```diff
@@ -187,15 +186,39 @@ export function isCreateAnnotationMessage(o: any): o is CreateAnnotationMessage
...
+  mode: "autodetect" | "fits" | "preloaded";
+  }
+
+/** Type guard function for CreateImageSetLayerMessage. */
+export function isCreateImageSetLayerMessage(o: any): o is CreateImageSetLayerMessage {  // eslint-disable-line @typescript-eslint/no-explicit-any
+  return typeof o.event === "string" &&
+    o.event == "image_layer_create" &&
+    typeof o.id === "string" &&
+    typeof o.url === "string" &&
+    typeof o.mode  === "string";
```
pkgw, line 212: In order to avoid unsafety in the type system, here we need to check whether `mode` is explicitly one of the allowed string values (`"autodetect"` etc.), and not just any string.

pkgw (commented):
Thanks, this is awesome! I really like how this PR is looking. Two small comments below, but I think there's only one last big thing to think through: how we handle the naming of ImageSets using their URLs and/or ImageSetID fields.

The example that @Carifio24 identified shows two substantively different imagesets that have the same image-data URL, and I think that the correct behavior would be to continue treating them as distinct imagesets. Fortunately, we have the ImageSetID field. I was originally thinking that we might change this field from a `number` to a `string`, but if it's used inside the tile cache, that's probably unwise. So I would propose that we start deriving the ImageSetID from a hash of both the URL and the FileType, which breaks the degeneracy in Jon's example. If there turn out to be more degeneracies, we can incorporate additional fields as needed (while hoping that our hash function is sufficiently strong).

When it comes to `GetImageSetByUrl`, I think that it is sensible to have a policy that in the HiPS case we should be fetching image data (FITS, PNG, JPG) and not catalogs (TSV). So when loading WTML files, I think we need to add some logic to skip the latter. I think the sensible place do to this would be in `WWTControl.AddImageSet`. How does that sound?

(Sooner or later, we might want to start adding HiPS catalogs to a separate list of imagesets, rather than just ignoring them altogether. We'll want to make it possible to "register" HiPS catalogs with the engine and pull them up, similar to what is possible with imagesets.)

(Also, NB, `WWTControl.AddImageSet` and `ScriptInterface.AddImageSet` do fairly different things; it might be good to rename one of them.)

Carifio24:
As I take another look at these two, they actually don't have the same URL - note the "UHSDR1_PIX" vs "UHSDR1_CAT" in the middle there. The original problem I had been having was how to distinguish between these in the background image selector (since most of the data is the same), but since the TSVs aren't going to be included there, maybe this isn't a problem after all.

pkgw:
> As I take another look at these two, they actually don't have the same URL - note the "UHSDR1_PIX" vs "UHSDR1_CAT" in the middle there.

Whoa, good point! I think most of what I wrote above is still relevant but we might not need to change anything with the ImageSetID for now.

engine/wwtlib/Folder.cs
imbasimba: Whoops, some Swenglish tried to sneak in...

imbasimba:
I think we are safe keeping the ImageSetID as is for now. At least for 99.99% of cases.

But, speaking of this issue in regards to the other PR
https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/96 
Would it ever make sense to open the same FITS image twice with different parameters (cuts, or similar)

pkgw:
> Would it ever make sense to open the same FITS image twice with different parameters (cuts, or similar)

I don't think that's something that we need to worry about supporting, no.

@imbasimba I think this is ready to merge — were you planning on making any other updates?

imbasimba:
No more updates :)

PR #94, opened 2021-05-18
New roll API for OpenSpace integration

pkgw:
CC @imbasimba. I'm extracting this piece of #91 to help simplify the review process for the other parts.


pkgw:
Also CC @Carifio24: here's an example of a change that plumbs a new API all the way from the C# layer to the Research App layer. There are a lot of layers involved but I've convinced myself that That's Just How It Has To Be.

PR #97, opened 2021-05-26
Layer ordering

imbasimba:
Follow up to 
https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/91
with:

1. Set layer order
2. Goto option exposed in the create imageset event


engine-helpers/src/index.ts
```diff
@@ -297,7 +297,7 @@ export interface LoadFitsLayerOptions {
 
   /** Whether to seek the view to the positon of the FITS file on the sky,
    * if/when it successfully loads. */
-  gotoTarget: boolean;
+  goto: boolean;
```
pkgw, line 300: Note that renaming these fields is an API break, although in this case I think it is acceptable.

engine/wwtlib/ScriptInterface.cs
```diff
@@ -446,6 +446,32 @@ private static ImageSetLayer AddFitsLayer(string url, string name, bool gotoTarg
             return imagesetLayer;
         }
 
+        public void SetImageSetLayerOrder(string id, int order)
+        {
+            foreach (string referenceFrame in LayerManager.AllMaps.Keys)
+            {
+                Layer layerToMode = GetLayerFromReferenceFrame(id, referenceFrame);
```
pkgw, line 453: I think that the lookup that's happening here might be inconsistent with the other APIs that modify layers. Those all find layers using TypeScript along these lines:

```ts
const layer = this.lm.get_layerList()[options.id];
if (layer && layer instanceof ImageSetLayer) {
   ...
}
```

This JS ends up referencing `LayerManager.LayerList`, which is a map from layer GUID to layer object. For consistency, this function should also match on the GUID stringification, not the Name.

engine/wwtlib/ScriptInterface.cs
```diff
@@ -446,6 +446,32 @@ private static ImageSetLayer AddFitsLayer(string url, string name, bool gotoTarg
...
+                
+                if(order >= 0 && layerToMode != null)
+                {
+                    LayerManager.AllMaps[referenceFrame].Layers.Remove(layerToMode);
+                    LayerManager.AllMaps[referenceFrame].Layers.Insert(order, layerToMode);
```
pkgw, line 458: What happens if `order` is bigger than `LayerManager.AllMaps[referenceFrame].Layers.Length`?

research-app-messages/src/classic_pywwt.ts
```diff
@@ -209,6 +212,7 @@ export function isCreateImageSetLayerMessage(o: any): o is CreateImageSetLayerMe
     o.event == "image_layer_create" &&
     typeof o.id === "string" &&
     typeof o.url === "string" &&
+    typeof o.mode  === "string" &&
     (o.mode == "autodetect" || o.mode == "fits" || o.mode == "preloaded");
```
pkgw, line 216: I'm pretty sure that it's a bit overkill, but I think the type guard function here should validate that `goto` is either undefined or a boolean. Since all JS objects have "truthiness" other types can work, but I think it is better to err on the side of being rigorous about input checking.

pkgw (commented):
Looks good! Just a few comments.

research-app-messages/src/classic_pywwt.ts
imbasimba: I had it, but lost it :disappointed: https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/97/commits/7eb7edb7a2961e4ff82a682fb838772873c12636

engine-helpers/src/index.ts
imbasimba: It is already deprecated layer options though, so this API break has very low value. I'll just revert it and have the new name for the new layer option class

engine/wwtlib/ScriptInterface.cs
imbasimba: Great! This triggered me to a new level of understanding for the Research app. Please have a look at the new approach, which is more in line with other layer messages. 

Though I am not yet sure of the utility vs hassle of the version sequence number that some messages have though. Would you mind elaborating a bit on the reasoning behind that?

engine/wwtlib/ScriptInterface.cs
imbasimba: They end up at the end of the list - (It is converted to a javascript array splice operation)

I added an explanatory comment for any future person having the same reasonable question.

engine/wwtlib/ScriptInterface.cs
pkgw: In at least one of the older pywwt use cases (Qt embedded browser or Jupyter widget) I believe that messages were sometimes arriving out-of-order, which could lead to bad UX if you don't catch that. I would expect the Qt-embedding and the Jupyter-widget cases to actually guarantee message ordering (see below), but the sequence numbering was definitely added for a reason — @astrofrog might remember the details. As for our newer apps that use the `postMessage` API, [this exchange][so] is a little unclear but suggests that the spec doesn't demand it but current implementations preserve message ordering. So for the time being, I *think* that maybe we can actually expect messages to arrive in order, but I haven't stress-tested to confirm that that's the case.

[so]: https://stackoverflow.com/questions/7536901/does-window-postmessage-guarantee-the-order-of-events

I like the idea of developing this subsystem defensively, but if messages can arrive out-of-order they can probably also get dropped completely and/or duplicated, and we're *definitely* not going to handle those conditions well. In the current code I tried to implement something that would handle the existing pywwt scheme decently, in part just to sketch out how to approach that, but haven't tried to break it. For new APIs I think we want to go for basic robustness in the message processing, but not spend too much effort on building in precautions unless we're really sure that they'll actually be useful. The docs should probably say "If you're writing a client that sends messages to an app, make sure that your messages are transmitted over a reliable medium that preserves ordering."

(Appendix: I would think that the Qt embedded message passing should be robust since the browser JS is being controlled directly by the Python code ... but I have no actual evidence on way or the other. In the Jupyter-widget case, messages from Python to the user's browser frontend do cross the open network, but they're relayed over a WebSocket channel, and those [ought to be reliable at the frontend level][so2].)

[so2]: https://stackoverflow.com/questions/11804721/can-websocket-messages-arrive-out-of-order/11809503

pkgw:
I added a couple of cosmetic tweaks here myself, since it didn't seem worthwhile to add the latency of another back-and-forth about them.

PR #98, opened 2021-05-27
Updated colormaps to use full color range

imbasimba:


pkgw (commented):
Thanks — looks like the tests need updating.

PR #99, opened 2021-05-28
Add keyboard support for WWT viewer control

Carifio24:
Adds keyboard support for zooming, moving, and tilting the WWT viewer. The mapping between keys and events is based on KeyboardEvent.key, which [per MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) has support on all browsers.

### DIFF
--- a/engine-vuex/src/store.ts
+++ b/engine-vuex/src/store.ts
@@ -349,6 +349,20 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
   }
 
   @Mutation
+  move(obj: { x: number; y: number }): void {
+    if (Vue.$wwt.inst === null)
+      throw new Error('cannot move without linking to WWTInstance');
+    Vue.$wwt.inst.ctl.move(obj.x, obj.y);
+  }
+
+  @Mutation
+  tilt(obj: { x: number; y: number}): void {
+    if (Vue.$wwt.inst === null)
+      throw new Error('cannot tilt without linking to WWTInstance');
+    Vue.$wwt.inst.ctl._tilt(obj.x, obj.y);
+  }
+
+  @Mutation
   setTime(time: Date): void {
     if (Vue.$wwt.inst === null)
       throw new Error('cannot setTime without linking to WWTInstance');
--- a/engine-vuex/src/wwtaware.ts
+++ b/engine-vuex/src/wwtaware.ts
@@ -295,7 +295,7 @@ export class WWTAwareComponent extends Vue {
         "updateTableLayer",
         "zoom",
         "move",
-        "tilt"
+        "tilt",
       ]),
     };
   }
--- a/engine-vuex/src/wwtaware.ts
+++ b/engine-vuex/src/wwtaware.ts
@@ -294,6 +294,8 @@ export class WWTAwareComponent extends Vue {
         "toggleTourPlayPauseState",
         "updateTableLayer",
         "zoom",
+        "move",
+        "tilt"
       ]),
     };
   }
@@ -581,6 +583,12 @@ export class WWTAwareComponent extends Vue {
    * [[gotoRADecZoom]].
    */
   zoom!: (f: number) => void;
+
+  /** Moves the position of the view */
+  move!: ({ x, y }: { x: number; y: number }) => void;
+
+  /** Tilts the position of the view */
+  tilt!: ({ x, y }: { x: number; y: number }) => void;
 
   // Actions
 
--- a/research-app/src/App.vue
+++ b/research-app/src/App.vue
@@ -427,30 +427,58 @@ class AnnotationMessageHandler {
   }
 }
 
+class KeyPressInfo {
+  code: string;
+  ctrl: boolean;
+  alt: boolean;
+  shift: boolean;
+  meta: boolean;
+
+  constructor(code: string, modifiers?: { ctrl?: boolean, alt?: boolean, shift?: boolean, meta?: boolean } ) {
+    this.code = code;
+    this.ctrl = modifiers?.ctrl ?? false;
+    this.alt = modifiers?.alt ?? false;
+    this.shift = modifiers?.shift ?? false;
+    this.meta = modifiers?.meta ?? false;
+  }
+
+  matches(event: KeyboardEvent): boolean {
+    return event.code === this.code 
+        && event.ctrlKey === this.ctrl
+        && event.altKey === this.alt
+        && event.shiftKey === this.shift
+        && event.metaKey === this.meta;
+  }
+}
+
 /** This simple class encapsulates how we handle key bindings */
 class KeyboardControlSettings {
-  zoomIn: string;
-  zoomOut: string;
-  moveUp: string;
-  moveDown: string;
-  moveLeft: string;
-  moveRight: string;
-  tiltCW: string;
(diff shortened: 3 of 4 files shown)


engine-vuex/src/store.ts
```diff
@@ -348,6 +348,20 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
     Vue.$wwt.inst.ctl.zoom(factor);
   }
 
+  @Mutation
+  move(obj: { x: number; y: number }): void {
+    if (Vue.$wwt.inst === null)
+      throw new Error('cannot move without linking to WWTInstance');
+    Vue.$wwt.inst.ctl.move(obj.x, obj.y);
```
pkgw, line 355: I think this PR is missing a change to `engine/src/index.d.ts` to expose the `move` and `_tilt` methods?

Note that `_tilt` is so named because it's marked as `private` in the C# — and it would actually be somewhat undesirable to make it public, because (e.g.) pywwt relies on the existing, underscore-prefixed name. The joys of maintaining API compatibility.

engine-vuex/src/wwtaware.ts
```diff
@@ -294,6 +294,8 @@ export class WWTAwareComponent extends Vue {
         "toggleTourPlayPauseState",
         "updateTableLayer",
         "zoom",
+        "move",
+        "tilt"
```
pkgw, line 298: Very minor stylistic thing: I think it is good to get into the habit of using trailing commas whenever possible, because it makes copy-pasting easier and reduces the noise on diffs,

research-app/src/App.vue
```diff
@@ -427,12 +427,54 @@ class AnnotationMessageHandler {
...
+    zoomIn = "ArrowUp",
+    zoomOut = "ArrowDown",
+    moveUp = "w",
+    moveDown = "s",
+    moveLeft = "d",
+    moveRight = "a",
+    tiltCW = "ArrowLeft",
+    tiltCCW = "ArrowRight",
+    moveAmount = 20,
+    tiltAmount = 20
```
pkgw, line 453: This is good but these keybindings are different than the ones that pywwt has been using. We should make these the same as the ones defined here:

https://github.com/WorldWideTelescope/pywwt/blob/master/pywwt/nbextension/static/wwt.html#L122-L187

If I'm remembering the numeric keycodes correctly, the definitions are:

- Zoom in/out are z/x or pgup/pgdn (my parity might be wrong here)
- Directional navigation is i/j/k/l or arrow keys
- Tilt is directional with Alt held down
  - I feel so out-of-touch — do Macs have Alt keys these days? If not we might need to provide an alternative
- There is a built-in delay effect that's lengthened by holding down Shift — I'm not sure what that accomplishes

research-app/src/App.vue
```diff
@@ -461,6 +503,34 @@ export default class App extends WWTAwareComponent {
         this.onMessage(event.data);
       }
     }, false);
+
+    // Handling key presses
+    this._keyboardControlSettings = new KeyboardControlSettings({});
+    this._keyListener = function(e: KeyboardEvent) {
+      const kcs = this._keyboardControlSettings;
+      if (kcs == null || kcs == undefined) { return; }
```
pkgw, line 511: Another stylistic thing: I would like to avoid one-line branches in control flow, i.e. `if` statements with their bodies all one one line. It makes it harder to catch the structure of control flow when scanning the code at a high level.

research-app/src/App.vue
```diff
@@ -461,6 +503,34 @@ export default class App extends WWTAwareComponent {
...
+      const movementAmount = kcs.moveAmount;
+      const tiltAmount = kcs.tiltAmount;
+      if (e.key === kcs.zoomIn) {
```
pkgw, line 514: If there are going to be multiple keys that trigger the same action (e.g. I/J/K/L and arrow keys for panning), then this part of things might need to be restructured. But this is one of those situations where there's no 100% elegant way to do things, so adding fields like `kcs.zoomInA` and `kcs.zoomInB` or whatever would be totally fine.

engine-vuex/src/wwtaware.ts
```diff
@@ -582,6 +584,12 @@ export class WWTAwareComponent extends Vue {
    */
   zoom!: (f: number) => void;
 
+  /** Moves the position of the view */
```
pkgw, line 587: If possible it would be good to document the units of the `x` and `y` values here ... but they might not be well-defined. Barring that, some sort of documentation of "reasonable" values might be helpful for future users of these APIs.

pkgw (commented):
Awesome! I have a few comments, nothing profound.

engine-vuex/src/wwtaware.ts
pkgw: (I didn't intentionally end that comments with a comma but it's so apropos that I'll leave it there!)

Carifio24: I'd never considered the effect that this would have on diffs- I'll add in the comma

research-app/src/App.vue
Carifio24: Makes sense - I'll change the keybindings to match pywwt.

My Macbook (Air 2015) doesn't have an Alt key - just Fn, Control, Option, and Command. The Mac keyboards on the Apple store seem to have the same layout, so it doesn't seem to have changed much. An alternative is a good idea, but I'm not sure which makes the most sense - maybe option?

Do we want the Shift delay to be implemented?

research-app/src/App.vue
Carifio24: I guess another option would be to make the keybinding fields in KeyboardControlSettings into arrays of strings, and then the check could be something like `kcs.zoomIn.contains(e.key)`

engine-vuex/src/store.ts
Carifio24: Yep! I will add in the definitions to the .d.ts files.

research-app/src/App.vue
pkgw: From playing around with pywwt, I can't figure out any way that the Shift delay helps — I think we can just ignore that.

As for modifiers, there are also complications about which modifiers are intercepted by browsers and/or the OS, and which aren't. A lot of websites seem to use Ctrl-Alt? I bet there's a good explanation somewhere out there about what modifiers you can safely hope to use in the web context ...

Carifio24: Ok, sounds good on the Shift delay - one less this to worry about. I'll take a look into modifier keys. Hopefully handling these isn't too painful.

I was thinking a bit more about the KeyboardControlSettings structure, and if we want to have multi-key controls (i.e. Alt + letter), together with the possibility to have multiple key options for each action, I think I'll need to juice it up a little. I have an idea on how to do that which might not be too messy.

pkgw:
So, I superseded this with #102 to make a few additions of my own — normally, I could push them onto the branch used in this PR, but I think that GitHub didn't let me do it in this case since the PR is coming from the `master` branch and not a feature branch.

Anyway, nice work @Carifio24! It was easy to build on what you did and add a few extra handy keybindings, which is a good sign about the approach here. With lots more keybindings it would probably be better not to have to add a different `keydown` event listener for each individual event that we handle ... but the scaling there might actually not be an issue at all. And for now it's OK.

PR #101, opened 2021-05-30
Sub optimal browser warning

imbasimba:
This is a suggestion on how to nudge people (probably mostly Safari users) to either enable WebGL 2.0 flag or switch to a browser that supports WebGL 2.0.

Since everything should still technically be working in WebGL 1.0, I did not want a to invasive message. Therefore the message is hidden after ~15 seconds. 

This also was a mini excersise to understand the WWTEngineVuexState & reactive properties, but maybe wwtShouldShowBrowserWarning is slightly misplaced there. Let me know what you think.


pkgw (commented):
Thanks — interesting approach here! I had an idea about generalizing this functionality, which maybe will come in handy for other warnings, or maybe is just making things too complicated.

pkgw:
@imbasimba Thanks, that's the right prioritization! I've pushed a change to resolve the merge conflict.

PR #103, opened 2021-06-02
Background selection in research app

Carifio24:
Added a background selection UI to the research app. Also added an array of available ImagesetInfo objects (these contain an ImageSet's URL, name, description, and type) to the Vuex store, for use in this background selection.


engine-vuex/src/store.ts
```diff
@@ -62,10 +63,28 @@ export class WWTGlobalState {
   }
 }
 
+export class ImagesetInfo {
```
pkgw, line 66: The rest of the file doesn't do a great job of this, but we should try to include docstring comments in new definitions such as this structure and its fields.

engine-vuex/src/wwtaware.ts
```diff
@@ -34,7 +34,8 @@ import {
   GotoRADecZoomParams,
   LoadTourParams,
   LoadImageCollectionParams,
-  WWTEngineVuexState
+  WWTEngineVuexState,
+  ImagesetInfo
```
pkgw, line 38: Another small stylistic point — for lists like these I try to maintain alphabetical ordering. In general, if there's any kind of collection of things in a source file, I like to alphabetize unless there's some other ordering or grouping that should be applied.

Also, trailing comma.

research-app/src/App.vue
```diff
@@ -46,9 +67,14 @@
 <script lang="ts">
 import * as moment from "moment";
 import * as screenfull from "screenfull";
+import vSelect from 'vue-select';
+import Vue from 'vue';
+import 'vue-select/dist/vue-select.css';
 import { Component, Prop, Watch } from "vue-property-decorator";
 import { fmtDegLat, fmtDegLon, fmtHours } from "@wwtelescope/astro";
 
+Vue.component("v-select", vSelect);
```
pkgw, line 76: I've been putting these calls in the `main.ts` file. It would be good to keep them together there, if there's not a specific reason for doing things differently here.

research-app/src/App.vue
```diff
@@ -624,6 +652,12 @@ export default class App extends WWTAwareComponent {
       screenfull.on('change', this.onFullscreenEvent);
     }
 
+    (async() => {
+      this.loadImageCollection({ url: this.hipsUrl, loadChildFolders: true });
+    })();
+
+    
+
```
pkgw, line 660: Can this just be a call to `loadImageCollection` without the async wrapper? I don't get any compiler errors if I do that ...

Also, I like to be very generous with my vertical spacing, but I generally keep spaces to one blank line at most within a function. (These comments are indicating that we should look into setting up automatic code formatting once we're past this deadline — this isn't stuff that needs to be managed manually.)

research-app/src/App.vue
```diff
@@ -910,6 +944,12 @@ export default class App extends WWTAwareComponent {
 
   // Background / foreground imagesets
 
+  get curAvailableImagesets() {
+    if (this.wwtAvailableImagesets == null)
+      return [];
+    return this.wwtAvailableImagesets.filter(info => info.type != ImageSetType.panorama);
```
pkgw, line 950: The way this filtering should work should be to only select those imagesets whose `type` is equal to `this.wwtRenderType`. The "render type" parameter basically defines the mode that the engine is currently in. Right now, for example, the first few entries in the menu ("Jupiter", "Moon", "Sun", etc.) are planetary maps. Changing the background imageset from one type to another is actually how you tell the engine to switch rendering modes — if you select Jupiter, you'll suddenly switch to the planet-viewer mode, rather than the sky-viewer mode. So showing imagesets of different types isn't a problem per se, but in terms of the UX I think we want to limit the dropdown to only show backgrounds that agree with the current rendering mode.

Also, this should filter out the "progressive HiPS catalogs", which can be identified because they have a FileType containing the text "tsv" (i.e., tab-separated values). I guess that means that the ImageSetInfo struct is going to have to record that piece of information as well — we're going to want to isolate those items for the source-catalog overlay functionality.

research-app/src/App.vue
```diff
@@ -1066,7 +1122,7 @@ body {
 #tools {
   position: absolute;
   z-index: 10;
-  bottom: 3rem;
+  top: 4rem;
```
pkgw, line 1125: I think it might be good to give this a top margin of `0.5rem`, or whatever the separation is for the coordinate readout and hamburger menu? Then everything will be nicely aligned at the top of the UI.

pkgw (commented):
This looks good! There are a few minor points to clean up but I think this is basically doing everything right.

Besides the comments below, also note that your PR deletes like 22,000 lines from the `package-lock.json` file here. That's a bit of a red flag, and is why I'm constantly using `git status` and `git show HEAD` to check that the changes that Git thinks I'm making line up with what I think I'm doing. What's going on here is that you seem to be using a slightly older version of `npm`, which uses a "version 1" lockfile format instead of "version 2". We don't want to be flip-flopping the lockfile back and forth from V1 to V2 formats, so you might look into upgrading — I use [nvm][nvm] to manage my Node/NPM versioning.

[nvm]: https://github.com/nvm-sh/nvm

pkgw (approved):
Looks great now!

And at some point we might want to go back and update the HiPS WTML URL, but the nice thing about being a web app is that we can change that later without it being much of a drag ...

PR #105, opened 2021-06-03
Fixups

pkgw:
- Unintentional API break
- Update Cranko internal versioning requirements

The API break manifested as the webclient "imagery" selector being empty. Here's the webclient code that relies on the old name:

https://github.com/WorldWideTelescope/wwt-web-client/blob/0177832de60978790287cfa346e845bcf2c153c0/controllers/MainController.js#L203

There are deployed forks of the webclient out there, so it will be easier to undo the API break than to try to update them all.

### DIFF
--- a/engine-vuex/package.json
+++ b/engine-vuex/package.json
@@ -19,8 +19,8 @@
   },
   "homepage": "https://worldwidetelescope.org/home/",
   "internalDepVersions": {
-    "@wwtelescope/engine": "68f56f0585354ec471ffe8efd2a8f6236ad82c27",
-    "@wwtelescope/engine-helpers": "68f56f0585354ec471ffe8efd2a8f6236ad82c27"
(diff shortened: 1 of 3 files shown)


Carifio24:
The web client works fine on my end using the wwtsdk.js created from this commit, so this looks good to me.

imbasimba:
Thanks for fixing this!

Hmm... I think my intention was to not expose the setter to keep the system marginally more safe, and make clients use the new GetImageSets() instead. 

Anyway, if there are forks floting around out there, it is probably best to keep the variable public.

pkgw:
Yes, I think the motivation was a good one, and we probably could get away with such a change in most cases. If there were only one version of the webclient to maintain I'd probably have just fixed it instead, but this seemed like the best approach given that there's at least one more version that I'm aware of.

PR #108, opened 2021-06-04
Fixed opacity in FitsShader

imbasimba:


pkgw (approved):
I haven't tested locally myself but this seems sufficiently straightforward that I'm not worried.

PR #109, opened 2021-06-04
Adjust z-indexes to fix #107 

### LINKED ISSUE #107: research-app: "tools" bar prevents access to hamburger menu
I realized that now that we've moved the toolbar to the top of the viewport, there's a UI issue. The toolbar is positioned with some nested divs like so:

```css
#tools {
  position: absolute;
  z-index: 10;
  top: 0.5rem;
  left: 50%;
 
  .tool-container {
    position: relative;
    left: -50%;
  }
}
```

On my Chrome, the outer `#tools` div ends up spanning the right half of the viewport, while the inner `.tool-container` div is nicely centered. But because the outer div has a high z-index, it covers up the hamburger menu on the right even though it is invisible over there.

To be honest I forget why I did the positioning this way. There are lots of ways to center things horizontally in CSS.

This problem might be solved simple by putting the `z-index` setting inside the `.tool-container` rather than the outer `#tools` but there might be a good reason that it was placed on the outside anyway.

Carifio24:
This fixes the issue of the controls being blocked, as the `.tool-container` won't overlap the controls unless the screen is small. I need to think about the CSS to handle small screens (in which case the selection element itself will overlap the controls), but that's for another time.

### DIFF
--- a/research-app/src/App.vue
+++ b/research-app/src/App.vue
@@ -1132,7 +1132,6 @@ body {
 
 #tools {
   position: absolute;
-  z-index: 10;
   top: 0.5rem;
   left: 50%;
   color: #FFF;
@@ -1140,6 +1139,7 @@ body {
   .tool-container {
     position: relative;
     left: -50%;
+    z-index: 10;
   }
 
   .opacity-range {


pkgw:
Superseded by #110 which also takes a whack at the width issue.

PR #111, opened 2021-06-09
Undefined check for ServiceWorker

imbasimba:
OpenSpace encountered an intreresting issue when deploying the research app to an http server.
See browser console here:
http://www.winterway.eu/wwt-http-problem/index.html?origin=http://www.winterway.eu

The short story is that ServiceWorker is undefined if a user connects with http instead of https, as you can see by connecting to the same code with https:
https://www.winterway.eu/wwt-http-problem/index.html?origin=https://www.winterway.eu

I tested this PR here, and it seems to be working fine on http:
http://www.winterway.eu/wwt/index.html?origin=http://www.winterway.eu

(ServiceWorker is defined on http conenction if the host is localhost, that's why at least I haven't seen the issue before)

@pkgw Do you remember the reasoning behind the `!instanceof MessagePort && !instanceof ServiceWorker`? By looking at the ts code it looks like a simple `event.source instanceof Window` would be enough, but maybe I am missing something?

### DIFF
--- a/research-app/src/App.vue
+++ b/research-app/src/App.vue
@@ -663,8 +663,9 @@ export default class App extends WWTAwareComponent {
         // destinations, but let's start simple. TypeScript currently requires
         // us to do an odd type guard here.
         if (this.statusMessageDestination === null) {
-          if (!(event.source instanceof MessagePort) && !(event.source instanceof ServiceWorker)) {
-            this.statusMessageDestination = event.source;
+          //ServiceWorker is only defined on https connections and localhost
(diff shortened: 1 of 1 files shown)


pkgw:
The reason is alluded to in the comment a few lines up: for whatever reason, TypeScript seems to require that the type guard be written in that form ... I copied it from somewhere else but I forget exactly where. When I tried the `instanceof Window` I got a compiler error, if I recall correctly. Perhaps that's been tidied up, though?

Anyway, good sleuthing and thanks!

PR #114, opened 2021-06-11
SkyImage non-fits fix

imbasimba:
A slightly more elegant solution is coming with the tiled fits feature


pkgw:
Thanks so much! Seems to work in my local test.

PR #115, opened 2021-06-11
Tiled FITS support + Merged SkyImage and TangentTile

imbasimba:
Engine support of tiled FITS! 🎆 

Test it:

1. https://winterway.eu/wwt-tiled-fits/index.html?origin=https://winterway.eu
2. postMessage({event: "load_image_collection", url: "https://winterway.eu/wwt-tiled-fits/index_rel.wtml"})
3. postMessage({event: "image_layer_create", id:"tiled-fits", url: "https://winterway.eu/wwt-tiled-fits/{1}/{3}/{3}_{2}.fits", mode:"preloaded", goto:true})
4. postMessage({event: "image_layer_stretch", id:"tiled-fits", stretch: 0, version: 1, vmin: 0, vmax: 1.01})

You can load the orignial fits with:
postMessage({event: "image_layer_create", id:"single-fits", url: "https://winterway.eu/wwt-tiled-fits/IRAC.fits", mode:"fits", goto:true})

The deployed version at https://winterway.eu/wwt-tiled-fits has two differences as compared to if you run this yourself with a TOASTY generated imageset. 1: the index.wtml is manually changed. 2:  all FITS are flipped (incorrectly) on the y axis to demonstrate the tiled FITS feature. TOASTY has a little mixup in the tile splitting/axis flipping, so the quick solution for demo purposes is to flip all FITS in the engine. In the next step I will look into the TOASTY issues.

Things to be developed for TOASTY:
1. Calculate min & max (possibly also suggested cuts)
2. TOASTY FITS tiling/y-axis mixup. 
3. Tiled FITS is shifted 1 pixel to the left compared to loading the original single FITS. Probably an off-by-one error in TOASTY, but I have not fully excluded a possible error in the engine.
4. Fix the WTML entries. The index_rel.wtml has been changed manually in a few ways to make it work

A nice next step would also be to adjust the zoom when loading and imageset and goto=true

### DIFF
--- a/engine/wwtlib/Graphics/Shaders.cs
+++ b/engine/wwtlib/Graphics/Shaders.cs
@@ -1816,8 +1816,11 @@ namespace wwtlib
                 uniform int scaleType;
                 uniform float opacity;
                 
-                bool isNaN(float value){
-                    return !(value < 0.0 || 0.0 < value || value == 0.0);
+                bool isNaN(float value) {
+                    // See https://stackoverflow.com/questions/9446888/best-way-to-detect-nans-in-opengl-shaders
+                    // PKGW also finds that we need `value != value` on his Dell laptop running
+                    // Chrome on Linux.
+                    return (value != value) || !(value < 0.0 || 0.0 < value || value == 0.0);
                 }
 
                 void main(void) {
@@ -1848,9 +1851,7 @@ namespace wwtlib
                         vec4 colorFromColorMapper = texture(colorSampler, vec2(physicalValue, 0.5));
                         fragmentColor = vec4(colorFromColorMapper.rgb, opacity);
                     }
-
                 }
-
                 ";
 
             String vertexShaderText =
--- a/engine/wwtlib/HealpixTile.cs
+++ b/engine/wwtlib/HealpixTile.cs
@@ -23,7 +23,6 @@ namespace wwtlib
         private bool subDivided = false;
         private readonly List<List<string>> catalogRows = new List<List<string>>();
         private WebFile catalogData;
-        private FitsImageTile fitsImage;
         private static readonly Matrix3d galacticMatrix = Matrix3d.Create(
                     -0.0548755604024359, -0.4838350155267381, -0.873437090247923, 0,
                     -0.8676661489811610, 0.4559837762325372, -0.1980763734646737, 0,
@@ -419,28 +418,6 @@ namespace wwtlib
             return true;
         }
 
-        public override void RenderPart(RenderContext renderContext, int part, double opacity, bool combine)
-        {
-            if (fitsImage == null)
-            {
-                TileShader.Use(renderContext, VertexBuffer, GetIndexBuffer(part, accomidation), texture2d, (float)opacity, false);
-            }
-            else
-            {
-                ColorMapContainer.BindColorMapTexture(PrepDevice, dataset.FitsProperties.ColorMapName);
-                FitsShader.Min = (float)dataset.FitsProperties.LowerCut;
-                FitsShader.Max = (float)dataset.FitsProperties.UpperCut;
-                FitsShader.ContainsBlanks = dataset.FitsProperties.ContainsBlanks;
-                FitsShader.BlankValue = (float)dataset.FitsProperties.BlankValue;
-                FitsShader.BZero = (float)dataset.FitsProperties.BZero;
-                FitsShader.BScale = (float)dataset.FitsProperties.BScale;
-                FitsShader.ScaleType = (int)dataset.FitsProperties.ScaleType;
-                FitsShader.TransparentBlack = dataset.FitsProperties.TransparentBlack;
-                FitsShader.Use(renderContext, VertexBuffer, GetIndexBuffer(part, accomidation), texture2d, (float)opacity, false);
-            }
-            renderContext.gl.drawElements(GL.TRIANGLES, TriangleCount * 3, GL.UNSIGNED_SHORT, 0);
-        }
-
         public void DrawCatalogTile(RenderContext renderContext, double opacity)
         {
             RenderedGeneration = CurrentRenderGeneration;
@@ -693,7 +670,7 @@ namespace wwtlib
                         });
                     } else
                     {
-                        FitsImageJs image = FitsImageJs.CreateHipsTile(dataset, URL, delegate (WcsImage wcsImage)
+                        FitsImageJs image = FitsImageJs.CreateTiledFits(dataset, URL, delegate (WcsImage wcsImage)
                         {
                             texReady = true;
                             Downloading = false;
--- a/engine/wwtlib/Imageset.cs
+++ b/engine/wwtlib/Imageset.cs
@@ -51,7 +51,7 @@ namespace wwtlib
                     }
                 case ProjectionType.SkyImage:
                     {
-                        return SkyImageTile.Create(level, x, y, imageset, parent);
+                        return new SkyImageTile(level, x, y, imageset, parent);
                     }
                 case ProjectionType.Plotted:
                     {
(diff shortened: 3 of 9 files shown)


pkgw:
Awesome!

Just a quick FYI: the ImageSet XML contains a "bottomsUp" field that basically sets the image parity, in the way that FITS and regular bitmap images are opposites of each other. I *know* that toasty and my Python modules don't deal with bottomsUp well. Using it properly is probably the best way to deal with the FITS y-axis inversion business.

pkgw:
I love how few lines of code are involved in turning this on!

imbasimba:
Ah, yes that bottomsUp/topDown I've seen in many places. Indeed, using it properly may be the best solution, I'll keep it in mind for the future

pkgw:
Ah, so I was having trouble finding the documentation that I had written because of a fairly esoteric bug that snuck into the doc-processing framework that made most of it all disappear! That's now fixed, so here's what we've currently got:

https://wwt-data-formats.readthedocs.io/en/latest/api/wwt_data_formats.imageset.ImageSet.html#wwt_data_formats.imageset.ImageSet

The different attributes of the class map more-or-less directly onto XML attributes and child elements.

pkgw:
Well, OK, my initial investigation seems to indicate that for tiled RGB images, BottomsUp just ... isn't implemented. With both the Windows and web clients, if I take a tiled image and flip that boolean in the WTML, it disappears. For non-tiled images (projection=SkyImage), that works fine, parity-flipping the image coordinate system.

imbasimba:
Following our discussion on Monday, I had a look at how the offsets are used in the engine, and I believe it is correct as it is right now. SkyImage is using offset as pixels, while TangentTile (tiled imageset) is interpreting offset as degrees.
TOASTY seems to always pop out the pixel offset version. This could be the source of the small small offset in position, as I have overridden the incorrect TOASTY pixel offset to 0. With a specific offset in degrees (hopefully the correctly calculated one), the positions will be overlapping perfectly.
posDiff

pkgw:
OK, I've been working on getting a handle of how different-parity images are handled throughout the stack. Here are some notes ...

Parity is really about the relationship between the image data coordinate system (thinking of X and Y pixel indices as real-numbered coordinates) and the screen/world coordinate system. JPEGs, as typically rendered, are "negative" ("top-down") parity: if the X axis of the image pixel coordinate system increases to the right of the screen, its Y axis increases downward. Note however that in WCS transformation matrices, there's a sign flip because the X axis of the RA/Dec coordinate system increases to the *left*: so in a negative-parity image, the WCS CD matrix has a *positive* determinant.

FITS images by default are the opposite: if the positive X axis of the image pixel coordinate system points to the right of the screen, its Y axis points upward. This is "positive" parity or WWT "bottoms-up". The WCS CD matrix of such an image has a *negative* determinant.

WCS specifications can express images of either parity. It's conventional for FITS to have positive parity, but not required. As of WorldWideTelescope/wwt_data_formats#37 I believe that the low-level `wwt_data_formats` code can now properly import and export WCS settings of both parities.

In the current support for (non-FITS) tiled images, the parity has to be negative. If I take a sample image and just flip its bottoms-up setting, it stops rendering.

If you have a 2D pixel buffer and WCS data, you can flip their parity without changing the appearance of the image. If you're flipping along the Y axis in Python:

```python
data = data[::-1]  # vertically flip data

wcs_headers['CD1_2'] *= -1
wcs_headers['CD2_2'] *= -1
wcs_headers['CRPIX2'] = data.shape[0] + 1 - wcs_headers['CRPIX2']  # plus one since FITS headers use 1-based indices
```

So if we have positive-parity FITS data (the common case) we can flip them to negative-parity.

So now I think the task is to understand how the current FITS rendering code deals with image parity, and whether it pays attention to the BottomsUp setting or what. To be able to reuse the tiled image rendering code, we probably need to make sure that the FITS renderer can properly handle negative parity (BottomsUp = false), and when we need to update toasty to make sure that when it tiles FITS it makes sure that the data are in negative-parity form.

pkgw:
I've just pushed an update to this branch to catch it up to `master`.

pkgw:
Note to self: to test HiPS FITS support and make sure that's still working:

1. Create a demo WTML file as below. We need a custom ImageSet entry that says that only FITS is supported, because otherwise the RGB image formats will be preferred. We also need to use a hash fragment in the URL to get past the engine's uniquifying code.
2. Serve the research app in the engine repo: `npm run serve-research`
3. Serve the custom WTML locally with CORS: `npx http-server --cors .` (we assume this ends up on port 8081)
4. Open the research app up in a mode that accepts commands: navigate Chrome to `http://localhost:8080/?origin=http://localhost:8080`
5. Manually command it to load the custom WTML: `postMessage({event: "load_image_collection", url: "http://localhost:8081/pkgw.wtml"});`
6. Use the background chooser to load `pkgw vizMine`
7. The network traffic analyzer should confirm that FITS files are being downloaded, and the display should look correct.

Sample WTML, saved as `pkgw.wtml`:

```
<?xml version="1.0"?>
<Folder MSRCommunityId="0" MSRComponentId="0" Permission="0" Name="PKGW" Group="Explorer"
  Searchable="False" Type="Sky">
  <VersionDependent>false</VersionDependent>
  <ImageSet DemUrl="" MSRCommunityId="0" MSRComponentId="0" Permission="0" Generic="False"
  DataSetType="Sky" BandPass="Visible"
  Url="http://alasky.u-strasbg.fr/ancillary/vizier/vizMine/Norder{0}/Dir{1}/Npix{2}#pkgwunique"
  TileLevels="3" WidthFactor="1" Sparse="False" Rotation="0" QuadTreeMap="0123"
  Projection="Healpix" Name="pkgw vizMine fits only" FileType="fits" CenterY="0" CenterX="0"
  BottomsUp="False" StockSet="False" ElevationModel="False" OffsetX="0" OffsetY="0"
  BaseTileLevel="0" BaseDegreesPerTile="180" ReferenceFrame="Sky" MeanRadius="1">
    <Credits />
    <CreditsUrl />
    <ThumbnailUrl>http://alasky.u-strasbg.fr/ancillary/vizier/vizMine/preview.jpg</ThumbnailUrl>
  </ImageSet>
</Folder>
```
Key customizations are in `Url` and `FileType`.

pkgw:
OK! I've been spending some time looking into the current state of affairs with FITS studies with this PR.

I can now get them to appear correctly with a bit of hacking. The way that I'm thinking about things is:

1. The coordinate system of the tiled imageset has to be negative parity — JPEG-like, BottomsUp=False. Otherwise things just won't render.
2. FITS files currently always render positive parity. I.e., within an individual images, FITS renders with a bottoms-up data layout.
3. A tiled FITS will render correctly if you do all of the data processing making sure that everything is in a negative-parity coordinate frame, and then just vertically flip the 2D data buffers for the individual FITS tile files to bring them into a bottoms-up layout.

So, consider a 2×2 quartet of FITS files that are part of a study tiling. With zero rotation, the X=0 Y=0 tile is on the top-left, X=1 Y=0 is on top right, etc — the tiling coordinate system is negative parity. But within each individual FITS tile, the data layout is bottoms-up — the x=0 y=0 pixel of the X=0 Y=0 tile is found at that tile's bottom left (middle left of the overall image), its x=0 y=255 pixel is at its top left, etc.

In order to cascade these four tiles into a higher-level tile, you can think of it two ways:

1. Load the four tiles. Vertically flip their individual data buffers to get them into top-down layout. Collect them into one 512×512 buffer in a negative-parity sense: X=0 Y=0 goes at the top-left. Downsample it. Vertically flip *that* buffer to write it out into the higher-level FITS file with bottoms-up layout again.
2. Load the four tiles. Collect them into one 512×512 buffer in a positive-parity sense: X=0 Y=0 goes at the bottom-left. Downsample it. Write out the buffer into the higher-level FITS file.

The second one sounds like it's going to be more efficient, doesn't it!

pkgw:
I believe that WorldWideTelescope/toasty#52 gets it so that we can generate correct tiled FITS datasets that render properly given this PR!

pkgw:
I've been putting this through its paces and it all seems to work as far as I can tell. Let's go!

PR #116, opened 2021-06-13
Simplified type guard

imbasimba:
Follow up to:
https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/111 

I just tested this simplified type guard and it seems to be working fine.

I have not seen any compiler warnings or errors (at least with my setup), so maybe the issue requiring the odd type guard has been fixed.

### DIFF
--- a/research-app/src/App.vue
+++ b/research-app/src/App.vue
@@ -660,12 +660,10 @@ export default class App extends WWTAwareComponent {
     window.addEventListener('message', (event) => {
       if (this.allowedOrigin !== null && event.origin == this.allowedOrigin) {
         // You could imagine wanting to send status updates to multiple
-        // destinations, but let's start simple. TypeScript currently requires
-        // us to do an odd type guard here.
+        // destinations, but let's start simple.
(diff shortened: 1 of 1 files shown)


pkgw:
Here is the discussion that motivated the current approach:

https://stackoverflow.com/a/53064737/3760486

I think the issue is that the source can be either a `Window` or a `WindowProxy`, and that for some reason one couldn't test directly for those two cases. But the comment is fairly old, so indeed maybe this has been cleaned up ... I will investigate a bit more.

pkgw:
I wonder if the reported issue, with `instanceof Window` unexpectedly returning false, was a manifestation of this:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_context_e.g._frames_or_windows

Anyway, things seem to work with the patch for me, and without a clear explanation of how things might go wrong, I'm OK with tidying this up. If something goes kablooie, we can fix it.

imbasimba:
Interesting. I'll look into this in the future

PR #117, opened 2021-06-17
Add support for panorama images to the embed app

Carifio24:
Adds support for panorama images in the embed app. If the given image in the WTML is a panorama, it's set as both the foreground and background images, since a foreground image is required for SetupForImagesetOptions.


pkgw (approved):
Looks good, works in my testing!

PR #118, opened 2021-06-20
Posting image_collection_loaded event on folder load complete

imbasimba:
Doing something more generic is also an idea, like this for example:
`this.statusMessageDestination.postMessage({event: msg.event + "_completed", originalMessage: msg}, this.allowedOrigin);`
Then the same style could be used for all async events, where the client may want to know when something has completed.

### DIFF
--- a/research-app-messages/src/classic_pywwt.ts
+++ b/research-app-messages/src/classic_pywwt.ts
@@ -32,6 +32,7 @@
  *
  * ### Image Sets
  *
+ * - [[ImageCollectionLoadedMessage]]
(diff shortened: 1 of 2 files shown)


pkgw:
I think we want to do this on a case-by-case basis for now since it might be helpful for the response message to include extra information for the client. In this particular case you could potentially send summary information about the loaded imagesets, for example, although we can start simple with just the notification.

We should also define TypeScript types for the response messages, primarily so that we can document their shapes in the API documentation.

Build failure seems to be related to some null checks that need to be added.

PR #119, opened 2021-06-25
Research app HiPS catalogs

Carifio24:
This PR adds UI functionality for displaying and managing HiPS catalogs in the research app, including the ability to control color and brightness of the display objects. A few notes:

- I added a Vuex module to store state specific to the research app (currently it's only the list of the HiPS catalogs that the user has active). As part of this, I changed the namespace of the engine-vuex store to be "wwt-engine" and let the new module be "wwt-research", since that seemed clearer to me.
- In the engine's index.d.ts, there's typo in the declaration for the Layer's function to set opacity - it's declared as `set_opaciy`. I don't think anything uses this (and if it does, it won't work, since it's not actually wired to anything), but in the interest of safety I left it there and added a new `set_opacity` declaration.
- If a HiPS catalog is removed and then re-added, the engine seems to remember the color it had used to display this catalog previously. I thought it felt more natural for the layer to start fresh if the catalog is re-added, so I have the research app reset the color using the engine's `addCatalogHipsByNameWithCallback`, but if anyone disagrees I'm fine with changing that behavior.


engine-vuex/src/store.ts
```diff
@@ -334,11 +335,11 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
...
-      && (Date.now() - this.timeAtStartup) < 15000;
-    if (this.showWebGl2Warning != showWebGl2Warning) {
-      this.showWebGl2Warning = showWebGl2Warning;
-    }
+    // const showWebGl2Warning = !wwt.si.isUsingWebGl2()
+    //   && (Date.now() - this.timeAtStartup) < 15000;
+    // if (this.showWebGl2Warning != showWebGl2Warning) {
+    //   this.showWebGl2Warning = showWebGl2Warning;
+    // }
   }
```
Carifio24, line 343: I commented this out because it was incessantly giving an error in the console on my machine. I assume that shouldn't be the case, though?

imbasimba:
> * If a HiPS catalog is removed and then re-added, the engine seems to remember the color it had used to display this catalog previously.

Indeed the HiPS properties & settings are stored on the imageset, and are therefore remembered.

engine-vuex/src/store.ts
imbasimba: Interesting... You don't happen to use Safari or some old version of a browser, do you?

The intention is to show a 15 second warning to users with browsers not supporting WebGL 2.

Carifio24:
Ah, I hadn't realized that information was stored on the imageset itself. That's good to know, thanks.

engine-vuex/src/store.ts
Carifio24: Turns out this was all on my end - I was using a bit out-of-date version of the `scriptsharp` artifact . Using the newest version, I don't get this error at all.

engine/src/index.d.ts
```diff
@@ -600,6 +600,7 @@ export class Layer {
   set_name(v: string): string;
   get_opacity(): number;
   set_opaciy(v: number): number;
+  set_opacity(v: number): number;
```
pkgw, line 603: I'm extremely confident that the misspelled API has never been used, since I would definitely have noticed the mistake, so I'm OK with removing it.

research-app/src/App.vue
```diff
@@ -52,10 +67,37 @@
                   label="name"
                   placeholder="Background"
                   >
+                  <template #option="option">
+                    <h4 style="margin:0">{{option.name}}</h4>
+                    <em style="margin:0; font-size:small;">{{option.description}}</em>
```
pkgw, line 72: FWIW, I believe that it's generally discouraged to include styling via the `style=` HTML attribute, even though it can get annoying to create whole CSS classes for small fixups such as these. See e.g. [the MDN docs][mdn] on `style`.

[mdn]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/style

research-app/src/App.vue
```diff
@@ -76,6 +118,8 @@ import * as screenfull from "screenfull";
 import 'vue-select/dist/vue-select.css';
 import { Component, Prop, Watch } from "vue-property-decorator";
 import { fmtDegLat, fmtDegLon, fmtHours } from "@wwtelescope/astro";
+import WWTResearchAppModule from "./store";
+import { mapMutations, mapState } from "vuex";
```
pkgw, line 122: For sequences of imports like these, I generally try to sort them in the following rough categories:

- Common external libraries
- Application-specific external libraries
- Internal dependencies

So I'd put the vuex import farther up, here.

research-app/src/CatalogItem.vue
```diff
@@ -0,0 +1,223 @@
...
+    @Prop({required: true}) catalog!: ImagesetInfo;
+    @Prop({required: false, default: Color.fromArgb(1, 255, 255, 255)}) defaultColor!: Color;
+    hasFocus = false;
+    isSelected = false;
+    visible = true;
+    color = new Color();
+    
+    beforeCreate(): void {
+      const wwtEngineNamespace = "wwt-engine";
+      const wwtResearchAppNamespace = "wwt-research";
```
pkgw, line 74: I see these constants have been duplicated a few times here and there. It might make sense to centralize them in a module somewhere, and import them.

research-app/src/TransitionExpand.vue
```diff
@@ -0,0 +1,93 @@
+<template>
+  <!-- This component is based on the code from https://markus.oberlehner.net/blog/transition-to-height-auto-with-vue/
```
pkgw, line 2: Thank you for "citing" where these things come from, it's very helpful!

research-app/src/TransitionExpand.vue
```diff
@@ -0,0 +1,93 @@
...
+.expand-leave-active {
+  transition: height 0.2s ease-in-out;
+  overflow: hidden;
+}
+
+.expand-enter,
+.expand-leave-to {
+  height: 0;
+}
+</style>
```
pkgw, line 93: This little icon here means that this file does not end with a newline character. "Best practice" (in Unix-land at least) is to include one; most edtiors have an option to enforce that there always is one.

research-app/src/store.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    @Mutation
+    removeResearchAppCatalogHips(catalog: ImagesetInfo) {
+        const index = this.hipsCatalogs.indexOf(catalog);
+        if (index >= 0) {
+            this.hipsCatalogs.splice(index, 1);
+        }
+    }
+
+}
```
pkgw, line 27: Another missing final newline.

pkgw (commented):
This is fantastic! The implementation pretty much looks great — I only have a few very small comments.

From my testing, I think I see a few areas to work on, but nothing that I think has to be addressed before merging this PR. For future polish though:

- I think we can tidy up the appearance of the coordinates/layers overlay a bit
- It would be nice to wire up the catalogs to automatically rotate through a nice color palette, as we've discussed
- Add a point-size scaler to the catalog UI controls
- Change the default marker for catalog items to be an X or O or square or something, since the default gaussian shape is too hard to distinguish from the underlying star imagery

The last item might be a bit of work to implement, since I believe that the engine has some limitations about the marker types it currently supports, but I think it will be pretty important for usability.

Once again, these are areas for future work but I think we should merge in this PR before tackling them, since it definitely provides a solid foundation on which to build these improvements. Great work Jon!

research-app/src/TransitionExpand.vue
Carifio24: Yep, there's a setting for that in VS Code - shouldn't be a problem in the future.

research-app/src/CatalogItem.vue
Carifio24: I agree. My initial thought was to put them in `main.ts`, but that would require the components to import from `main.ts` and create a circular dependency as far as imports go. I've moved these to `namespaces.ts` in `research-app/src`, but if we end up using a similar structure elsewhere, we can/should move it further up the repository tree.

PR #123, opened 2021-07-17
research-app: Source selection for HiPS catalogs

Carifio24:
This PR adds functionality for research app users to select sources from HiPS catalogs for further examination by clicking on them. Every time the mouse moves, the app finds the closest source (within a zoom-dependent distance threshold). I'm using the `getTableDataInView` method from the `SpreadsheetLayer` that the HiPS catalog lives in, which as far as I can see is the best way to get that data out.

* I'm not a huge fan of how the research app store is set up from a data structure standpoint, but I've stuck with arrays for the time being since we can use `Vue.set` to keep things reactive. This is useful since it allows options like wiring particular entries of the visibility array into corresponding catalog components in the display panel. For the future, I want to look into whether there's a way to achieve this in Vue 2 with a data structure that would make more sense (i.e. a `Map` for the HiPS catalogs, and a `Set` for the sources).

* The name generation for sources still needs some work. I've put in the identifiers for a few star catalogs and will generate names based on those when applicable, but any suggestions/discussion on this front are definitely welcome.

### DIFF
--- a/engine-vuex/src/store.ts
+++ b/engine-vuex/src/store.ts
@@ -14,21 +14,23 @@ import {
 
 import {
   Annotation,
-  Color,
   Constellations,
   EngineSetting,
   Folder,
   Guid,
   Imageset,
   ImageSetLayer,
+  InViewReturnMessage,
   SpreadSheetLayer,
   WWTControl,
 } from "@wwtelescope/engine";
 
 import {
-  ApplyFitsLayerSettingsOptions,
+  AddCatalogHipsByNameOptions,
   AddImageSetLayerOptions,
+  ApplyFitsLayerSettingsOptions,
   ApplyTableLayerSettingsOptions,
+  GetCatalogHipsDataInViewOptions,
   GotoTargetOptions,
   LoadFitsLayerOptions,
   SetFitsLayerColormapOptions,
@@ -180,7 +182,7 @@ export interface WWTEngineVuexState {
   tourTimecode: number;
 
   showWebGl2Warning: boolean;
-  
+
   /** The current zoom level of the view, in degrees.
    *
    * The zoom level is the angular height of the viewport, times size.
@@ -438,11 +440,18 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
     Vue.$wwt.inst.setupForImageset(options);
   }
 
-  @Mutation
-  addCatalogHipsByName(name: string): void {
+  @Action({ rawError: true })
+  addCatalogHipsByName(options: AddCatalogHipsByNameOptions): Promise<Imageset> {
     if (Vue.$wwt.inst == null)
       throw new Error('cannot addCatalogHipsByName without linking to WWTInstance');
-    Vue.$wwt.inst.ctl.addCatalogHipsByName(name);
+    return Vue.$wwt.inst.addCatalogHipsByName(options);
+  }
+
+  @Action({ rawError: true })
+  getCatalogHipsDataInView(options: GetCatalogHipsDataInViewOptions): Promise<InViewReturnMessage> {
+    if (Vue.$wwt.inst == null)
+      throw new Error('cannot getCatalogHipsDataInView without linking to WWTInstance');
+    return Vue.$wwt.inst.getCatalogHipsDataInView(options);
   }
 
   @Mutation
@@ -450,30 +459,6 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
     if (Vue.$wwt.inst == null)
       throw new Error('cannot removeCatalogHipsByName without linking to WWTInstance');
     Vue.$wwt.inst.ctl.removeCatalogHipsByName(name);
-  }
-
-  @Mutation
-  addCatalogHipsByNameWithCallback(args: { name: string; callback: () => void }): void {
-    if (Vue.$wwt.inst == null)
-      throw new Error('cannot addCatalogHipsByNameWithCallback without linking to WWTInstance');
-    Vue.$wwt.inst.ctl.addCatalogHipsByNameWithCallback(args.name, args.callback);
-  }
-
-  @Mutation
-  setCatalogHipsColorByName(args: { name: string; color: Color }): void {
-    if (Vue.$wwt.inst == null)
-      throw new Error('cannot setCatalogHipsColorByName without linking to WWTInstance');
-    const layer = Vue.$wwt.inst.lm.get_layerList()[args.name];
-    layer.set_color(args.color);
-    layer.set_opacity(args.color.a);
-  }
-
-  @Mutation
-  setCatalogHipsOpacityByName(args: { name: string; opacity: number }): void {
-    if (Vue.$wwt.inst == null)
-      throw new Error('cannot setCatalogHipsOpacityByName without linking to WWTInstance');
-    const layer = Vue.$wwt.inst.lm.get_layerList()[args.name];
-    layer.set_opacity(args.opacity);
   }
 
   @Mutation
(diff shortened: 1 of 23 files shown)


pkgw:
OK, I've just updated your branch with an update that attempts to merge in all of the changes that were on `master`. Now for the actual review ...

research-app/src/App.vue
```diff
@@ -1648,13 +1737,85 @@ export default class App extends WWTAwareComponent {
     return imagesets.filter(iset => iset.name.toLowerCase().includes(searchText) || iset.description.toLowerCase().includes(searchText));
   }
 
+  // RA, Dec in radians
+  distance(pt1: AngleCoordinates, pt2: AngleCoordinates): number {
```
pkgw, line 1741: This function would be a great candidate for moving into the `astro` package, which aims to bundle together generic astronomy-ish routines that don't depend on WWT-specific types.

research-app/src/App.vue
```diff
@@ -1648,13 +1737,85 @@ export default class App extends WWTAwareComponent {
...
+  }
+
+  closestInView(target: AngleCoordinates, threshold?: number): Source | null {
+    let minDist = Infinity;
+    let closestPt = null;
+
+    //const pointsInView = this.dataInView();
+    //const result = this.closest(target, pointsInView);
+    // const closePt = result.pt
+    // const minDist = result.dist
```
pkgw, line 1759: We can probably just remove this commented-out code.

research-app/src/App.vue
```diff
@@ -1648,13 +1737,85 @@ export default class App extends WWTAwareComponent {
...
   data() {
     return {
       showPopover: false,
       showLayers: true,
+      showSources: true,
     }
   }
 
+  updateDistanceThreshold = debounce((app: App, newZoom: number) => {
+    app.distanceThreshold = 0.00002 * newZoom;
```
pkgw, line 1812: Where is this magic factor coming from? A quick explanatory comment would be helpful.

research-app/src/App.vue
```diff
@@ -1978,4 +2139,19 @@ ul.tool-menu {
...
 
+.pointer {
+  cursor: pointer;
+}
+
+/**
+This makes the last element of the last list item in the
+display panel have the rounded bottom edge
+The alternative to this is to have Vue bind a class to the last element
+Since I'm using
```
pkgw, line 2150: Sentence fragment in the comment here.

research-app/src/SourceItem.vue
```diff
@@ -0,0 +1,224 @@
...
+      return fmtHours(this.source.ra);
+    }
+
+    get decStr() {
+      return fmtDegLat(this.source.dec);
+    }
+
+    get searchRadius() {
+      return 2 * this.wwtDegZoom / 360; // In arcminutes
+    }
```
pkgw, line 100: This should be clamped to never go below a certain minimum value, say 10 arcseconds. When I was playing around I got a case where a SIMBAD search failed because the search radius was like sub-arcsecond, and in realistic use cases the online databases aren't going to necessarily agree with the catalogs to that level precision.

pkgw (commented):
Very cool! I have a few ideas about how to polish up the UX, but this is looking like a really nice foundation upon which to build. I have a couple of suggestions for rearranging things a bit but no major comments.

People have been requesting this functionality for *years* so it's great to see it come together!

engine-vuex/src/store.ts
```diff
@@ -279,6 +280,60 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
     }
   }
 
+  get findRADecForScreenPoint() {
+    return function (pt: { x: number; y: number }): { ra: number; dec: number } {
+      if (Vue.$wwt.inst === null)
+        throw new Error('cannot findRADecForScreenPoint without linking to WWTInstance');
+      const coords = Vue.$wwt.inst.ctl.getCoordinatesForScreenPoint(pt.x, pt.y);
+      return { ra: (15 * coords.x + 720) % 360, dec : coords.y };
```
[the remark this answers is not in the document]
Carifio24: Yeah, this will make sure that the RA value is positive. I actually stole this from the [FinderScope](https://github.com/WorldWideTelescope/wwt-web-client/blob/master/factories/FinderScope.js#L27) in the web client.

research-app/src/App.vue
Carifio24: I don't really have any reasoning for this factor - when I was testing out this functionality, I ended up landing on this value because I thought it 'felt good' as far as how close the cursor needed to get to a source before the cursor changed.

I was thinking - with the circle annotations that we're now going to have around the sources, maybe it would be better to use those distances as the threshold for each point?

research-app/src/SourceItem.vue
Carifio24: Makes sense. I've set 10 arcseconds (= 1/6 arcminutes) as the minimum value.

research-app/src/App.vue
Carifio24: I honestly have no clue what I was going to write, so I just removed that line.

research-app/src/App.vue
```diff
@@ -769,28 +790,42 @@ export default class App extends WWTAwareComponent {
 
   defaultColor = Color.fromArgb(1, 255, 255, 255);
   wwtComponentNamespace = wwtEngineNamespace;
-
+  lastClosePt: Source | null = null;
+  distanceThreshold = 0.01;
   hideAllChrome = false;
   hipsUrl = "http://www.worldwidetelescope.org/wwtweb/catalog.aspx?W=hips"; // Temporary
+  drag = false;
```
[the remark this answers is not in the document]
Carifio24: I've changed it to `isMouseMoving`, but I'm open to other suggestions.

engine-vuex/src/store.ts
```diff
@@ -279,6 +280,60 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
...
+      if (Vue.$wwt.inst === null)
+        throw new Error('cannot get currentHipsCatalogData without linking to WWTInstance');
+      const layer = Vue.$wwt.inst.lm.get_layerList()[name];
+      if (layer != null && layer instanceof SpreadSheetLayer) {
+        return layer.getTableDataInView();
+      } else {
+        return ''
+      }
+    }
+  }
```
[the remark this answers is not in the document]
Carifio24: Yeah, that's a good point re: the tediousness of adding getters. As you suggested, I've made one getter that returns the `SpreadSheetLayer`.

research-app/src/App.vue
```diff
@@ -740,6 +756,11 @@ class KeyboardControlSettings {
   }
 }
 
+interface AngleCoordinates {
+  ra: number;
+  dec: number;
```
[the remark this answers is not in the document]
Carifio24: I was using that indexer at some point while working on this, but yeah, it didn't end up being needed.

engine-vuex/src/wwtaware.ts
```diff
@@ -451,6 +458,15 @@ export class WWTAwareComponent extends Vue {
    */
   lookupImageset!: (_n: string) => Imageset | null;
 
+  /** Get the constellation closest to the given point */
+  findConstellationForPoint!: (pt: {ra: number; dec: number}) => string;
+
+  /** Get the right ascension and declination, in degrees, for x, y coordinates on the screen */
```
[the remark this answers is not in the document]
Carifio24: This was something that I was experimenting with that I forgot to remove. I don't think it's actually used anywhere, so I think I'll just take it out. I believe three-letter all-caps abbreviations is the output format, though.

research-app/src/App.vue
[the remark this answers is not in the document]
pkgw: I like that! I feel like that's much clearer.

pkgw (approved):
OK, looking good — I think this will be ready to merge once we add a brief comment about the distance threshold conversion factor.

PR #127, opened 2021-07-30
Show default sky background menu when tours are paused

pkgw:
As requested by @astrodavid10. This has some correctness issues if the tour isn't in sky mode, but we have those issues anyway.

The only substantial code changes in this PR are in `embed/src/Embed.vue`. I also turned on auto-formatting in my VS Code which reformats a bunch of the files in a more standard way. I also fixed some linter and doc mistakes that had snuck in.


astrodavid10:
Yay, thanks!

pkgw:
OK, I've just cued up a new release of the embedder that should deploy this change to production over the next hour or so.

PR #129, opened 2021-08-06
engine/wwtlib/Layers/ImageSetLayer.cs: properly handle tiled-fits imagesets too

pkgw:
The test for fitsiness was only working for single FITS images.

#### Draft

Filing a PR to build the C# and see if this actually works ...


pkgw:
Cool, looks like this works!

PR #130, opened 2021-08-06
research-app: tighten up the panel presentation a bit

pkgw:
@Carifio24 Here are some suggested changes to try to tighten up the presentation of the research app panel:

- Give the panel a transparent background, as opposed to giving the whole thing a 60% opacity. The latter approach makes it so that nothing within the panel shows up at full intensity
- Tighten up the internal margins. `<p>` elements have to have their margins zeroed so as not to add undue vertical spacing
- Downscale the section headings ("Sources:" etc), using a little inline rule effect instead
- Ellipsize catalog and source names rather than letting them wrap
- Give them a `vertical-align: middle` to even out their presentation by a few pixels
- Kill redundant background settings in the catalog and source item components
- Add some extra `cursor: pointers` when hovering over catalog/source item labels (to expand the items) and the color selector
- Reduce internal padding in the detail rows

The catalog URL doesn't show up nicely in this version. It gets truncated, but the element width isn't limited so you don't get a nice ellipsis. We should probably add a "Copy URL to clipboard" button — I think ellipsization + that is nicer than showing the full URL.

Also, a lot of the layout here might benefit from using CSS "flexbox" layout. It's helpful for seemingly-simple cases like the icons that come after the catalog/source names. And I think it will probably help make the layout reflow more naturally for narrow layouts.

### DIFF
--- a/research-app/src/App.vue
+++ b/research-app/src/App.vue
@@ -16,7 +16,7 @@
       </transition>
       <div id="layers-container" v-if="haveLayers">
         <div class="display-section-header">
-          <label>Layers:</label>
+          <label>Layers</label>
         </div>
         <div v-if="showLayers">
           <catalog-item
@@ -29,7 +29,7 @@
       </div>
       <div id="sources-container" v-if="haveSources">
         <div class="display-section-header">
-          <label>Sources:</label>
+          <label>Sources</label>
         </div>
         <div v-if="showSources">
           <source-item
@@ -2051,7 +2051,7 @@ body {
 }
 
 #overlays {
-  margin: 10px;
+  margin: 5px;
 }
 
 #controls {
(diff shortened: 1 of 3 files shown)


pkgw:
@Carifio24 I had already made the dividers pure white, so now I've doubled the thickness to 2px.

Carifio24:
Your comment just now made me realize - I often use the Dark Reader extension on Chrome, and I had apparently forgotten to turn it off when I looked at this the other day. The effect on the rest of the panel is essentially nil, but it was turning the lines black instead of white (hence my difficulty seeing them). So either of the thicknesses is fine with me.

pkgw:
Hmm, interesting. That's the kind of thing we should be mindful of in the future (even if it's caused by an extension, it may be a relatively common phenomenon) but I think we're in good shape here, then.

PR #131, opened 2021-08-10
Work on reactive settings

pkgw:
@Carifio24 OK, here is what I think we need to do to be able to wire up the app UI to control the WWT state with proper "reactive" UI magic. How does this all look to you? I'd recommend you read the changes commit-by-commit — hopefully that will give a sense as to how the pieces fit together.

I also took some inspiration from [this post] for how to approach the general state storage.

[this post]: https://forum.vuejs.org/t/vuex-best-practices-for-complex-objects/10143/2

If we go with this approach, I can go through and flesh out the "settings" frameworks for the rest of the objects that need it — there are a lot of them, but it's all pretty rote work.

The bottom line is that I'm pretty sure that we need to do this mirroring of WWT objects for the Vue(x) reactive UI magic to work. It doesn't feel elegant but I think it's the most effective approach.

### DIFF
--- a/engine-helpers/src/index.ts
+++ b/engine-helpers/src/index.ts
@@ -6,9 +6,7 @@ import { D2H, R2D, R2H } from "@wwtelescope/astro";
 import {
   ImageSetType,
   isBaseEngineSetting,
-  isBaseLayerSetting,
   isBaseImageSetLayerSetting,
-  isBaseSpreadSheetLayerSetting,
   isBaseVoTableLayerSetting,
 } from "@wwtelescope/engine-types";
 
@@ -25,10 +23,8 @@ import {
   ImageSetLayer,
   ImageSetLayerSetting,
   InViewReturnMessage,
-  Layer,
   LayerManager,
   LayerManagerObject,
-  LayerSetting,
   Place,
   Poly,
   PolyAnnotationSetting,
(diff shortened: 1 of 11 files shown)


Carifio24:
Overall, I think this looks really good! I think doing things like this will make wiring up various settings to the UI straightforward, albeit a bit tedious at first, like you said.

I did see one issue - I think `addCatalogHipsByName` in the engine Vuex store might need something akin to the `getIndex`-based checks in the `research-app` store - if a catalog that's already present is selected again, the color of the point markers is reset to white (with the UI left unchanged).

pkgw:
@Carifio24 Ah, I see what you mean. I think that with the hashmap, we don't need to worry about duplication there ... but the `color` field in the CatalogItem now duplicates the actual layer `color` setting, which is why I think things are getting out of sync. Should be straightforward to deduplicate, I think.

pkgw:
All right, I think I've tied up the loose ends for this PR. In principle the app's catalog "visibility" state could/should be normalized to derive from the engine's knowledge of layer state, but we don't *have* to do that, I think.

If you re-add a HiPS catalog layer the color gets reset to white, I think because of the `onMounted` method for CatalogItem, but it should now at least be properly synchronized between the UI and the engine.

Carifio24:
Looks good to me! I'll try to think of a way to handle the color change/re-adding issue, but that can be for the future.

PR #132, opened 2021-08-11
Add a scale factor control for HiPS catalogs

pkgw:
I spent a surprisingly long time trying to figure out the UI here. I started with a slider, but I couldn't figure out a way to make that work given the large dynamic range needed by the scale-factor parameter. Then for numeric inputs, the basic web browser styling is dissatisfactory, but I couldn't find an powered-up Vue input widget with good TypeScript bindings. So I slightly rolled my own.

### DIFF
--- a/research-app/src/CatalogItem.vue
+++ b/research-app/src/CatalogItem.vue
@@ -36,10 +36,12 @@
           <span class="prompt">URL:</span
           ><span class="ellipsize">{{ catalog.url }}</span>
         </div>
+
         <div class="detail-row" v-if="catalog.description.length > 0">
           <span class="prompt">Description:</span
           ><span>{{ catalog.description }}</span>
         </div>
+
         <div class="detail-row">
           <span class="prompt">Color:</span>
           <v-popover class="circle-popover">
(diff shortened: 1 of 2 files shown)


Carifio24:
The only issue I can see is that the input seems a bit jumpy with regards to typing - if I try to enter a decimal, like 1.5, the input automatically guesses that I want 1.00 as soon as I hit 1, and I end up with "+1.00.5" in the input. I tried changing the input to use `v-model.lazy`, and that seemed more natural to me - that way the field only updates when the user presses enter or brings the focus elsewhere.

pkgw:
@Carifio24 Good idea. Are you able to commit that fix and push the update directly to my branch? I'm not sure about whether we have the GitHub permissions set up for that.

Carifio24:
@pkgw It doesn't look like I can. I'm guessing something would need to be adjusted in the repository settings?

pkgw:
I've tried futzing with the permissions — for the record, I gave the WWT developers team "maintain" privileges on the repo. That may help? But at any rate, I've pushed a `.lazy` change of my own — I think it's definitely a nicer UX that way.

PR #133, opened 2021-08-11
Research app selection messaging

Carifio24:
This PR adds a new message type, `SelectionStateMessage`, to `research-app-messages`, as well as functionality in the research app to broadcast these messages at appropriate times. Currently, the items that are broadcast on a change are the list of selected HiPS catalogs, the list of selected sources, and the most recent selected source.

I'm not sure if this is exactly how we want to handle the most recent selected source (which would be useful for e.g. `glue-wwt`). As things are now, every time a new source is selected, messages are broadcast for both the list of sources and the most recent source, which feels redundant. The alternative, I think, would be to only broadcast the source list message, and leave it on the user to identify whether the last source in the list is new or not.

### DIFF
--- a/research-app-messages/src/index.ts
+++ b/research-app-messages/src/index.ts
@@ -13,6 +13,14 @@ export {
   settings,
 };
 
+export interface Source {
+  ra: number;
+  dec: number;
+  name: string;
+  catalogName: string;
+  zoomDeg?: number;
+}
+
 /** Information about the current position of the WWT view.
  *
  * Frontends (e.g., [embedded WWT apps][embed]) periodically send this
@@ -211,11 +219,8 @@ export interface SelectionStateMessage {
   sessionId: string;
 
   /** The most recent source that was added to the selection list. */
-  mostRecentSource?: string;
-
-  /** The list of HiPS catalogs that are currently selected. */
-  selectedCatalogs?: string[];
+  mostRecentSource?: Source;
 
   /** The list of sources that are currently selected. */
-  selectedSources?: string[];
+  selectedSources?: Source[];
 }
--- a/research-app-messages/src/index.ts
+++ b/research-app-messages/src/index.ts
@@ -197,10 +197,10 @@ export function isPingPongMessage(o: any): o is PingPongMessage {  // eslint-dis
 
 export interface SelectionStateMessage {
 
-    /** The tag identifying this message type. */
-    type: "wwt_selection_state";
+  /** The tag identifying this message type. */
+  type: "wwt_selection_state";
 
-    /** An app/client session identifier.
+  /** An app/client session identifier.
    *
    * If a single client is communicating with multiple apps, it needs to be able
    * to tell which app is the source of any update messages. This session
@@ -218,14 +218,4 @@ export interface SelectionStateMessage {
 
   /** The list of sources that are currently selected. */
   selectedSources?: string[];
-}
-
-/** Type guard function for [[SelectionStateMessage]] */
-export function isSelectionStateMessage(o: any): o is SelectionStateMessage { // eslint-disable-line @typescript-eslint/no-explicit-any
-  return typeof o.type === "string" &&
-    o.type == "wwt_selection_state" &&
-    typeof o.threadId === "string" &&
-    (o.sessionId === undefined || typeof o.sessionId === "string") &&
-    (o.mostRecentSource === undefined || typeof o.mostRecentSource === "string") &&
-    (o.selectedSources === undefined || typeof o.selectedSources === "string");
 }
--- a/research-app-messages/src/index.ts
+++ b/research-app-messages/src/index.ts
@@ -180,3 +180,52 @@ export function isPingPongMessage(o: any): o is PingPongMessage {  // eslint-dis
     typeof o.threadId === "string" &&
     (o.sessionId === undefined || typeof o.sessionId === "string");
 }
+
+/** Information about the current state of source and catalog selection
+ * inside the WWT application
+ * 
+ * This message is broadcasted by the application whenever one of the user's selection
+ * options inside of the application is adjusted. The Vue listeners for the selected 
+ * catalogs and sources are deep listeners, meaning that this message will be broadcasted
+ * whenever a property of one of their items changes.
+ * 
+ * Not all fields of this message will always be present, depending on the
+ * nature of the event triggering the emission of this message. A message
+ * missing a particular field should be treated as conveying no information
+ * about the state described by that field.
+ */
+
+export interface SelectionStateMessage {
+
+    /** The tag identifying this message type. */
+    type: "wwt_selection_state";
+
+    /** An app/client session identifier.
+   *
+   * If a single client is communicating with multiple apps, it needs to be able
+   * to tell which app is the source of any update messages. This session
+   * identifier allows clients to do so. The default value is "default". But if
+   * a client sends a [[PingPongMessage]] with a customized ``sessionId`` field,
+   * that value will start appearing in these view state update messages.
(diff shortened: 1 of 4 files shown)


research-app-messages/src/index.ts
```diff
@@ -180,3 +180,52 @@ export function isPingPongMessage(o: any): o is PingPongMessage {  // eslint-dis
...
+
+    /** An app/client session identifier.
+   *
+   * If a single client is communicating with multiple apps, it needs to be able
+   * to tell which app is the source of any update messages. This session
+   * identifier allows clients to do so. The default value is "default". But if
+   * a client sends a [[PingPongMessage]] with a customized ``sessionId`` field,
+   * that value will start appearing in these view state update messages.
+   */
+  sessionId: string;
```
pkgw, line 211: The whitespace here has gotten wonky somehow.

research-app-messages/src/index.ts
```diff
@@ -180,3 +180,52 @@ export function isPingPongMessage(o: any): o is PingPongMessage {  // eslint-dis
...
+}
+
+/** Type guard function for [[SelectionStateMessage]] */
+export function isSelectionStateMessage(o: any): o is SelectionStateMessage { // eslint-disable-line @typescript-eslint/no-explicit-any
+  return typeof o.type === "string" &&
+    o.type == "wwt_selection_state" &&
+    typeof o.threadId === "string" &&
+    (o.sessionId === undefined || typeof o.sessionId === "string") &&
+    (o.mostRecentSource === undefined || typeof o.mostRecentSource === "string") &&
+    (o.selectedSources === undefined || typeof o.selectedSources === "string");
```
pkgw, line 230: There are some problems here — there's no `threadId` in the message and some of the types here should be array types. But overall, since this is an *outgoing* message, we don't have TypeScript code that needs to do rigorous checking of the message structure, so I think we should be able to skip this function altogether.

Carifio24: Yeah, various issues aside, you're right - we don't need this.

Carifio24:
I didn't have a particular use case in mind for updating the users on the active HiPS catalogs. I was thinking of 'selection' in the broader sense of 'things the user has chosen to display', so it felt natural to include there, but I'm fine with sticking just to sources.

For the source format, I think your suggestions of either using either the TSV-string or dictionary formats are probably the best bet. Either method will require the user to parse things like numerical values themselves. Like you said, the user should already be able to parse TSV output, so maybe we should go with that?

For columns, each source contains `ra` and `dec` columns which contain the respective value in radians (as well as `catalogName`, a string containing the name of the catalog) that are used for the source selection in the app, so I think we can get away without column hints, since those are guaranteed to be present.

pkgw:
@Carifio24 For the source data, it can just come down to whatever's more convenient to generate in the frontend. With the way that the source data ultimately come from the catalog data export, I think that would be TSV?

Carifio24:
@pkgw  Actually, the app stores each source as an object (whose values are all strings, with the exception of RA and Dec), so it's easier to just put the `Source` object(s) directly into the message.

From a `pywwt` point of view, this has the benefit that the received object is already a dictionary (I had expected that it would be a string representation of a dictionary, and so the user would need to parse it themselves, but that doesn't seem to be necessary).

research-app-messages/src/index.ts
```diff
@@ -180,3 +182,17 @@ export function isPingPongMessage(o: any): o is PingPongMessage {  // eslint-dis
...
+ * This message is broadcasted by the application whenever one of the user's selection
+ * options inside of the application is adjusted. The Vue listeners for the selected 
+ * catalogs and sources are deep listeners, meaning that this message will be broadcasted
+ * whenever a property of one of their items changes.
+ * 
+ * Not all fields of this message will always be present, depending on the
+ * nature of the event triggering the emission of this message. A message
+ * missing a particular field should be treated as conveying no information
+ * about the state described by that field.
+ */
```
pkgw, line 198: Er, I think this docstring needs to be relocated?

research-app-messages/src/selections.ts
```diff
@@ -0,0 +1,35 @@
...
+    /** The tag identifying this message type. */
+    type: "wwt_selection_state";
+  
+    /** An app/client session identifier.
+     *
+     * If a single client is communicating with multiple apps, it needs to be able
+     * to tell which app is the source of any update messages. This session
+     * identifier allows clients to do so. The default value is "default". But if
+     * a client sends a [[PingPongMessage]] with a customized ``sessionId`` field,
+     * that value will start appearing in these view state update messages.
```
pkgw, line 25: Hm, this docstring should probably just refer the reader to the description of `ViewStateMessage.sessionId` for its semantics. I see that I just copy-pasted the view-state doc into `ApplicationStateMessage.sessionId` as well, but it would be better to have a single source of truth for this kind of thing.

research-app-messages/src/selections.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Information about a selected source in the engine.
+ * The values for right ascension and declination are given in radians.
+ * `catalogName` gives the name of the associated HiPS catalog,
+ * while `name` gives a string name for the source.
+ */
+ export interface Source {
+    ra: number;
+    dec: number;
+    name: string;
+    catalogName: string;
```
pkgw, line 11: Here I think we need to declare a TypeScript indexer to indicate that there will be additional fields for the catalog columns, and mention that in the docstring.

It might be good to namespace the arbitrary catalog data more strongly (e.g., you can worry about potential issues about well-definedness if the catalog provides a column named `ra`, etc.) but we're mirroring the internal app data structures here, and I'd prefer not to spend too much time fiddling with those, and I think what we've got here will be fine.

pkgw (commented):
OK, a couple of fiddly bits about the interface definition and docs to tidy up, I think, but I think we've got the fundamentals down here.

research-app-messages/src/index.ts
Carifio24: Yep - this got left behind when I was moving things into `selections.ts`.

research-app-messages/src/selections.ts
Carifio24: That makes sense. I'll modify this description to reference `ViewStateMessage.sessionId`.

research-app-messages/src/selections.ts
Carifio24: This potential issue of well-definedness of name is also present in the app itself. I agree that we don't want to spend too much time messing with the data structure in the app, but I'm thinking we could go with something along the lines of:

```
export interface Source {
    ra: number;
    dec: number;
    name: string;
    catalogName: string;
    info: {
        [field: string]: string | undefined;
    }
}
```

where the `info` object contains all of the catalog data.

pkgw: Yeah, that would be the most sensible way to encapsulate, I think. Maybe name the field `catalogData` or something. If you think it will be straightforward to modify the app's Source object this way too, I do think it would be good for future-proofiness.

Carifio24:
So I modified the `Source` interface in the format we had discussed in both the messages and the research app itself.

While I was modifying the app's Source to use this format, I noticed an issue with the way that the `Source` objects were being constructed - basically, they had initially been created without a `name`, but TypeScript hadn't noticed because they were partially created from an unpacked object. To fix this, I added a `RawSourceInfo` interface, which is now how the app stores the closest source when the mouse is moving, before the user clicks on it. This then gets converted into a `Source` and added to the database when the user makes the selection. This change should also make `closestInView` ever-so-slightly more efficient.

pkgw (approved):
OK, I think that's everything!

PR #134, opened 2021-08-24
Work on reactive settings, FITS layer UI in the app

pkgw:
Set up more Vue(x) infrastructure to expose more settings reactively, this time with ImageSet layers.

**updated:** This is working towards a basic UI for dealing with FITS-like layers, including tiled FITS. I prototyped all of the functionality here but am now breaking it into a series of PRs to make it more reviewable.

### DIFF
--- a/engine-vuex/src/store.ts
+++ b/engine-vuex/src/store.ts
@@ -704,7 +704,7 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
   get imagesetStateForLayer() {
     return (guidtext: string): ImageSetLayerState | null => {
       if (Vue.$wwt.inst === null)
-        throw new Error('cannot get imagesetStateForLAyer without linking to WWTInstance');
+        throw new Error('cannot get imagesetStateForLayer without linking to WWTInstance');
       return this.imagesetLayers[guidtext] || null;
     }
   }
--- a/engine-vuex/src/store.ts
+++ b/engine-vuex/src/store.ts
@@ -758,8 +758,8 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
     Vue.$wwt.inst.stretchFitsLayer(options);
 
     // Update the reactive mirror.
-    const state = this.imagesetStateForLayer(options.id);
-    if (state !== null) {
+    const state = this.imagesetLayers[options.id];
+    if (state) {
       state.scaleType = options.stretch;
       state.vmin = options.vmin;
       state.vmax = options.vmax;
@@ -774,8 +774,8 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
     Vue.$wwt.inst.setFitsLayerColormap(options);
 
     // Update the reactive mirror.
-    const state = this.imagesetStateForLayer(options.id);
-    if (state !== null) {
+    const state = this.imagesetLayers[options.id];
+    if (state) {
       state.colormapName = options.name;
     }
   }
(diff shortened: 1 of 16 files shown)


pkgw:
OK, I think that all of the pieces are now in place! Next I'm going to work on breaking this into multiple PRs ... I'll start by updating this to stick closer to what the title says, and file new ones as we work through the review.

pkgw:
@Carifio24 OK, this PR is still pretty big, but now a bit more focused. Mainly defining new settings infrastructure to enable more reactive Vue(x) UIs linked to the WWT engine.

engine-vuex/src/store.ts
```diff
@@ -621,35 +697,55 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
...
   }
 
+  // Imageset layers, including FITS layers
+
+  imagesetLayers: { [guidtext: string]: ImageSetLayerState } = {};
+
+  get imagesetStateForLayer() {
+    return (guidtext: string): ImageSetLayerState | null => {
+      if (Vue.$wwt.inst === null)
+        throw new Error('cannot get imagesetStateForLAyer without linking to WWTInstance');
```
Carifio24, line 707: Typo ("LAyer") in the error message.

Carifio24:
I think the setup for the various settings types looks good! Should the console commands still work? They worked fine when I tried this out last week, but now I'm getting an error when I run the layer creation command:

```
index.umd.js?2adf:4835 Uncaught (in promise) TypeError: this.imagesetStateForLayer is not a function
    at Object.1aba.WWTEngineVuexModule.applyFitsLayerSettings (index.umd.js?2adf:4835)
    at Store.mutation (index.umd.js?2adf:784)
    at wrappedMutationHandler (vuex.esm.js?94ea:844)
    at commitIterator (vuex.esm.js?94ea:466)
    at Array.forEach (<anonymous>)
    at eval (vuex.esm.js?94ea:465)
    at Store._withCommit (vuex.esm.js?94ea:624)
    at Store.commit (vuex.esm.js?94ea:464)
    at Store.boundCommit [as commit] (vuex.esm.js?94ea:409)
    at Store.local.commit (vuex.esm.js?94ea:796)
```

pkgw:
Hmm. The console commands won't work, because this PR now only contains the first stage of the new functionality, but I don't think you should be seeing the error that you've recorded there. I think maybe an `npm run build` before the `npm run serve-research` should fix it?

Carifio24:
Hmm...I had started from a fresh clone of your branch before, so I had run `npm build` anyways, but just as a sanity check I re-grabbed `wwtlib.js` from last commit in this PR, and re-built, and I'm still getting the same error.

I think the problem might be accessing the Vuex getter inside a mutation. If so, we could replace the getter call with `this.imagesetLayers[guidtext] || null` - all the getter itself adds is the null check on the instance, which all of the mutations do themselves anyways.

pkgw:
@Carifio24 Ah, this could be helpful, because I know that there are other cases where I'm seeing similar errors that have something to do with that sort of thing. Is it a rule that mutations can't access getters? Can they invoke other mutations?

pkgw:
Anyway, a later commit made exactly the change you suggested, so I'll "backport" it to this PR.

Carifio24:
@pkgw: It seems that you can't call getters from mutations - most of the discussion I see around this point is from ~2017 (like [here](https://github.com/vuejs/vuex/issues/684)), but I think our issue indicates that that's still the case.

It seems that you can commit a mutation from another mutation (rather than calling it directly), but the recommended approach is to wrap a "multiple mutation" inside an action.

Carifio24:
That did it - the error is gone now. I think this first stage is good to go then.

PR #136, opened 2021-09-01
Initial testing

Carifio24:
So it took a bit longer than I expected, but here's the initial testing setup. The tests are written in [Nightwatch](https://nightwatchjs.org/). Currently there's only a basic set of tests for the research app. Obviously we may need to make additional adjustments for e.g. getting this set up on Azure.

This is all in the `tests` folder, which has the following contents

* Configuration files for running the test suite completely locally (`local.conf.js`), through BrowserStack on a local version of the app (`bslocal.conf.js`), and on BrowserStack (`nightwatch.conf.js`). The last one may need to be modified as we get this set up on Azure.
* A script for running BrowserStack on a local app (`scripts/bslocal.runner.js`).
* `tsconfig.json` and `package.json` files.
* A Nightwatch [Page Object](https://nightwatchjs.org/api/pageobject/) for the research app (`page_objects/researchApp.js`). Using a page object allows us to essentially define an API for what sections/elements exist on our page, as well as define commands. This way, if the selector we would use to access an element changes, we only need to modify the page object - the test only needs to be changed when the behavior changes. Currently the url of this object is set to `localhost:8080` - I'm not sure what the best thing to do will be when we're running it through Azure.
* The basic test of research app functionality (`research_app/basic_test.ts`). This runs through a few simple pieces of functionality of the website:
1. Testing that the WWT component loads
2. Verifying that everything displays correctly in the initial configuration
3. Opening the background/HiPS catalog choosers and counting the number of options
4. Selecting a HiPS catalog, and making sure that the corresponding layer item in the display panel has the correct name and can be expanded and deleted.

A few notes:

* The tests are written in TypeScript, and are then built into the `dist` directory. Nightwatch gets upset if it sees an `export` statement, so I've set it to ignore TypeScript and run the generated `.js` files.
* The TypeScript support in the tests themselves is a bit thin - Nightwatch creates the page object instance from the object in the `.js` file, and I'm still working on what exactly the TS should look like for our particular instance.
* There's an issue running a local version of the app on BrowserStack for Safari and Opera, which seems to be related to using `localhost`. I'm still looking into how exactly we want to solve this.


pkgw:
This is very slick! Now I'll work on the Azure Pipelines integration. Hopefully it won't take too many tries, but you kind of just have to keep on making and pushing small iterative commits until it all works ...

tests/nightwatch.conf.js
```diff
@@ -0,0 +1,285 @@
+// Autogenerated by Nightwatch
+// Refer to the online docs for more details: https://nightwatchjs.org/gettingstarted/configuration/
+const Services = {}; loadServices();
```
pkgw, line 3: Is this file used at all? It looks like the two local configurations derive from it?

pkgw (commented):
Looks great, and like I said earlier — very slick! I've tweaked the README a bit based on my experience with local testing and set up the CI with the best approach I've been able to figure out.

It would definitely be good to get Safari testing going. Any progress on that?

The other thing is that I think right now the tests are all run on Windows. Would be good to get other OS's, including Android and iOS. The research app UI is *not* going to work well on mobile, which is OK. But the embed *should* work well on mobile and hopefully we can set up some tests for that.

Carifio24:
As far as Safari goes, the BrowserStack FAQ has [this](https://www.browserstack.com/question/663) advice to offer. I'm not sure how helpful this is in an Azure context - namely, is it possible to get the IP address of whatever machine the build is running on?

I can add some other OSes to the config files. Like you said, Android and iOS will be nice to have, even if we aren't going to use them immediately.

tests/nightwatch.conf.js
Carifio24: This was automatically generated, and I used it as a source for a lot of the things in other two config files, but we ultimately don't need it.

pkgw: OK, then I think we should remove it to reduce the potential for confusion.

pkgw:
@Carifio24 For Safari, it depends on whether they need a globally visible, public IP address, or whether a private IP will do. The language of the docs isn't entirely clear to me on that. But if a private IP suffices, that's easy to obtain from the shell. It looks like the Azure runners have IP addresses in the 10.x space. (The `npm run serve-research` command prints some output that includes the local IP.)

Carifio24:
I've added in some additional OS environments to the BrowerStack configuration file. In the course of doing this, I converted the configuration files to TypeScript files - now everything that we run is in the `dist` folder. The `package.json` commands have been updated accordingly. In the course of doing this, I discovered that the TypeScript support for Nightwatch only covers up to version 1.3 (current is 1.7.8) - so I rolled a few interfaces of my own that will hopefully help keep things organized.

It seems that BrowserStack is failing to recognize Edge for some reason. The `bs-local` command that we're running on Azure works on my machine - I wouldn't think it would be any different on Azure, since it's being run through BrowserStack.

Carifio24:
@pkgw The issue with Edge not being recognized is resolved (the trick is to use 'MicrosoftEdge' as the browser name - clearly [this](https://www.browserstack.com/automate/capabilities), while still useful, is not completely up-to-date).

pkgw:
@Carifio24 Great! Are there any more changes you were planning to make to this PR?

Carifio24:
@pkgw Solving the Safari/localhost issue would be nice. But I would be fine pushing that into its own PR, since I have no idea how much effort it'll entail.

pkgw:
So if I add "Safari" to the current `bslocal.conf.ts` and run various Safari tests on BrowserStack, it seems to be working reliably for me. Is that not the case for you? I'm running from a Fedora Linux machine, although I wouldn't think that would matter.

Carifio24:
It's working for me as well. I would guess that whatever issue was there was got resolved by the updates to `bslocal.conf.ts`, but I'm not sure what change made it work.

Carifio24:
With Safari good to go (and added into the Azure pipeline), that's everything I was looking to do for this PR.

PR #137, opened 2021-09-01
FITS UI part 3: engine-vuex extensions

pkgw:
Here we add a few more hooks to the engine-vuex module to help the frontend know what the active layers are. The code that manages the spreadsheet layers should be updated to use the same infrastructure ­— the LayerMaps hierarchy is the correct source of truth for this information.


Carifio24 (approved):
Looks good to me!

PR #138, opened 2021-09-02
FITS UI part 4: the actual research app UI

pkgw:
Final pull request in the series. Adds the new UI, similar to the catalog and source UIs.

To demo the functionality:

1. Run research app
2. Use new Load WTML Collection tool and load: http://data1.wwtassets.org/packages/2021/08_hscfits/index.wtml
3. Use new Add Imagery as Layer tool and add "HSC-Sample"
4. Experiment with the new UI

CC @imbasimba for visibility.

### DIFF
--- a/engine-vuex/src/store.ts
+++ b/engine-vuex/src/store.ts
@@ -185,8 +185,7 @@ export class ImageSetLayerState {
  * normalized when expressed in Vuex, as in [this post]. The WWT engine
  * certainly does *not* express its state in such a manner.
  *
- * [this post]:
- * https://forum.vuejs.org/t/vuex-best-practices-for-complex-objects/10143/2
+ * [this post]: https://forum.vuejs.org/t/vuex-best-practices-for-complex-objects/10143/2
  *
  * The duplication of WWT's data structures is annoying, but the actual amount
  * of mirrored data isn't very big.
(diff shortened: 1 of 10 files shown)


pkgw:
Hmm, it looks like the interactive cutoff adjustment is not working well at all on Firefox — the view isn't updating with mouse movements the way it does on Chrome. It looks like I need to adjust the way that the code listens for mouse movements.

Carifio24:
Yeah, it's giving me issues as well - for me, the first cutoff that I adjust on Firefox seems to work correctly, but any subsequent times that I try to just the cutoff, the value only changes when my mouse is over the display panel.

pkgw:
OK, I think I fixed it now. Silly failure to clean up all of my event handlers (though I'm not sure why it wasn't causing issues on Chrome).

Carifio24:
I'm not seeing any issues on Firefox now. After using this a bit more, I have two suggestions/comments:

- I think it would be nice to give some sort of acknowledgement that a user's WTML file was successfully loaded (or wasn't, if that's the case)
- Do we want to display so many decimal points in the cutoff text fields? It seems like a lot to me, but I'm not very familiar with the context here

pkgw:
For the WTML loading, totally agree. I was just putting off implementing it since I thought it might be a bit of a pain. I'll look into it and try to get a sense as to how much work it might take.

For the cutoff precision, the current behavior is surely excessive but there could be cases where you need a high level of precision — depending on the data, the overall color scale range might need to be very small compared to the actual data values. We could try to get fancy and, like, adaptively scale the precision based on that ratio, but that seems like a fair amount of work for not a ton of payoff.

pkgw:
New sample file:

http://data1.wwtassets.org/packages/2021/09_phat_fits/index.wtml

This provides 55 GB of PHAT FITS data in two images

pkgw:
OK, final (?) tweaks pushed ...

PR #139, opened 2021-09-11
Fix for uninteractable area around notifications

imbasimba:
Easing myself into this again with a really simple PR. 

Basically there is always an area of 0.75rem height that cannot be interacted with since it is covered by vue-notification-group. When the notification is showing the area is even bigger making the control button and part of the zoom button.


pkgw:
Thanks!

PR #140, opened 2021-09-14
Transparent black disabled by deafault

imbasimba:
As discussed, the transparentBlack property is now off by default, just like the current pywwt behaviour.

Also two tiny changes

1. to open a help link in a new tab 
2. fixed unnecessarily wide display-panel for larger screens.


pkgw:
Great, thanks!

PR #142, opened 2021-09-15
Research app UI updates

Carifio24:
This PR contains a few small UI updates to the research app. I'm still working on adjusting the overall UI layout based in response to screen size, but I can see that requiring some discussion and iteration, so I'm thinking that can go in its own PR.

The changes in this PR are:

- Retain the `border-radius` of the bottom item in the display panel when it's highlighted. Currently, when the bottom item is highlighted, its bottom corners return to being rectangular. I suspect that the structure in the panel items in the panel changed at some point and the CSS hadn't been updated to match.
- Ellipsize the items in the dropdown selectors to avoid horizontal scrolling in the dropdowns, as well as the selected items (i.e. for the backgrounds and imagery layers).
- Change the container with the labels and buttons in the catalog item to use a flexbox-based layout. If we like this layout, I can change the other display panel items to use the same setup.

### DIFF
--- a/research-app/src/App.vue
+++ b/research-app/src/App.vue
@@ -179,8 +179,8 @@
             >
               <template #option="option">
                 <div class="item-option">
-                  <h4>{{ option.name }}</h4>
-                  <em>{{ option.description }}</em>
+                  <h4 class="ellipsize">{{ option.name }}</h4>
+                  <p class="ellipsize"><em>{{ option.description }}</em></p>
                 </div>
               </template>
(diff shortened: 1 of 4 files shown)


pkgw:
Looks like the test suite needs an update to track one of these changes?

Carifio24:
Yep - made a small DOM change that broke one of the page object selectors. All fixed now.

pkgw:
OK, this is all looking good to me!

PR #144, opened 2021-09-20
Research app display panel UI

Carifio24:
This PR changes the source entries in the display panel to use a flexbox-based layout to match the catalog and imageset containers. Additionally, using `ImagesetItem` as the base, I made the buttons have uniform size and spacing across all of the containers.

### DIFF
--- a/research-app/src/CatalogItem.vue
+++ b/research-app/src/CatalogItem.vue
@@ -1,11 +1,13 @@
 <template>
-  <div id="root-container">
+  <div
+    id="root-container"
+    @mouseenter="hasFocus = true"
+    @mouseleave="hasFocus = false"
+    @focus="hasFocus = true"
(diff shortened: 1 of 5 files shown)


Carifio24:
The BrowserStack test is failing because Safari is picking up an extra space at the end of an element's text content. All other browsers are fine. I'll look around and see if this is a known issue.

Carifio24:
The `gap: 2px` in the `#main-container` of each item seems to be what is adding the extra space in the `label` in Safari - removing this removes the space, and that was newly added to the catalog item in this PR.

To get around this, I've changed that particular test to match against a regex which allows trailing spaces. At the moment it's a `RegExp` literal defined in the section props. I want to clean up the selectors in the testing and add some tests for the WTML loading sometime soon; it would be good to automate that regex creation too.

PR #146, opened 2021-09-20
Allow UI manipulation and selection of general table layers

Carifio24:
This PR adds initial support for manipulating user-added table layers via the UI, and selecting sources present in these layers. The main idea is as follows:

- I added a `SpreadSheetLayerInfo` class, analogous to `ImagesetInfo` that we were using for the HiPS catalogs, which contains the relevant information for the display panel and research app store about a spreadsheet layer.
- `CatalogItem` has been renamed to `SpreadsheetItem` (and its `catalog` prop renamed to `layer`). The `layer` prop now has type `LayerInfo = ImagesetInfo | SpreadSheetLayerInfo`. Since these layer types, and the corresponding info classes, are pretty similar, a few `v-if` statements and some delegation in the delete method are enough to handle pretty much all of the differences. Since the engine-generated names are very non-user-friendly, I added the ability to edit the layer name of a general table layer, copying the setup from `SourceItem`.
- The research app store has been modified to use a key-based data structure for layers, with the key being the ID for a table layer, and the name for a HiPS catalog (the same keying system used by the engine). The data structure values now hold three pieces of information: the type of the layer (`hips` or `table`), whether or not the layer is visible, and the `LayerInfo` itself. 
- Since sources can now come from places other than HiPS catalogs, I replaced the `catalogName` in a `Source` with a `layer` field containing the `LayerInfo` of the layer from which the source came. Keeping a reference to the `LayerInfo` object in the `Source` will be useful for reactive stuff - for example, as things are now, if the user changes the name of a table layer, the panel item for all of the sources from that layer automatically update as well. This necessitated changing the `Source` interface, and adding some other interfaces, in `research-app-messages`.
- Finally, I added a couple of methods to the `engine-vuex` store to get the layer state for non-HiPS layers, which are largely just clones of the HiPS version.

I suspect there might be some iteration on this, but I think this general setup should make handling both types of layers not too painful.

### DIFF
--- a/engine-vuex/src/store.ts
+++ b/engine-vuex/src/store.ts
@@ -117,6 +117,35 @@ export class ImagesetInfo {
   }
 }
 
+export function isImageSetInfo(o: any): o is ImagesetInfo { // eslint-disable-line @typescript-eslint/no-explicit-any
+  return typeof o.url === 'string'
+      && typeof o.name === 'string'
+      && typeof o.type === 'number'
+      && typeof o.description === 'string'
+      && typeof o.extension === 'string';
+}
+
+export class SpreadSheetLayerInfo {
+  /* The user-facing name of the layer */
+  name: string;
+
+  /* The internal GUID of the layer */
+  id: string;
+
+  /* The reference frame in which the data are defined. */
+  referenceFrame: string;
+
+  setName(name: string) {
+    this.name = name;
+  }
+
+  constructor(id: string, referenceFrame: string, name?: string) {
+    this.id = id;
+    this.referenceFrame = referenceFrame;
+    this.name = name ?? id;
+  }
+}
+
 /** Information about an active imageset layer. */
 export class ImageSetLayerState {
   /** Layer parameters exposed in WWT's generic "settings" system.
@@ -1004,6 +1033,21 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
     }
   }
 
+  get spreadSheetLayerById() {
+    return function(id: string): SpreadSheetLayer | null {
+      if (Vue.$wwt.inst === null)
+        throw new Error('cannot get spreadSheetLayerById without linking to WWTInstance');
+
+      const layer = Vue.$wwt.inst.lm.get_layerList()[id];
+
+      if (layer !== null && layer instanceof SpreadSheetLayer) {
+        return layer;
+      } else {
+        return null;
+      }
+    }
+  }
+
   get spreadsheetStateForHipsCatalog() {
     return (name: string): SpreadSheetLayerSettingsInterfaceRO | null => {
       if (Vue.$wwt.inst === null)
@@ -1011,6 +1055,15 @@ export class WWTEngineVuexModule extends VuexModule implements WWTEngineVuexStat
 
       // NOTE: hack as described above -- the name is the GUID.
       return this.spreadSheetLayers[name] || null;
+    }
+  }
+
+  get spreadsheetState() {
+    return (id: string): SpreadSheetLayerSettingsInterfaceRO | null => {
+      if (Vue.$wwt.inst === null)
+        throw new Error('cannot get spreadsheetState without linking to WWTInstance');
+
+      return this.spreadSheetLayers[id] || null;
     }
   }
 
--- a/engine-vuex/src/store.ts
+++ b/engine-vuex/src/store.ts
@@ -138,6 +138,14 @@ export class SpreadSheetLayerInfo {
   }
 }
 
+/** This type defines the union of the various spreadsheet layers.
+ *
+ * Currently this includes standard spreadsheet layers, and HiPS catalog
+ * layers, which are a sort of hybrid between spreadsheet and imageset
+ * layers.
+ */
+export type CatalogLayerInfo = SpreadSheetLayerInfo | ImagesetInfo;
+
 /** Information about an active imageset layer. */
 export class ImageSetLayerState {
   /** Layer parameters exposed in WWT's generic "settings" system.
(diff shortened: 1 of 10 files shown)


engine-vuex/src/store.ts
```diff
@@ -117,6 +117,35 @@ export class ImagesetInfo {
   }
 }
 
+export function isImageSetInfo(o: any): o is ImagesetInfo { // eslint-disable-line @typescript-eslint/no-explicit-any
```
pkgw, line 120: I think that we don't need this function? I'm pretty sure that the relevant tests can be done with the TypeScript `instanceof` operator. The key difference versus some of our other code is that `ImagesetInfo` is a class, while the various `is*` functions are for interfaces, or classes defined in the ScriptSharp engine code where it's not clear that the `instanceof` operator will do the right thing. But here, I'm pretty sure that `instanceof` will meet our needs.

research-app-messages/src/selections.ts
```diff
@@ -1,3 +1,57 @@
+export enum ImageSetType {
```
pkgw, line 1: Hmmm. I think it would be good not to define ImageSetType as an enum here. The reason is that a big purpose of the research-app-messages module is to make it clear how the messages are expressed in a JSON serialization, and the use of an enum type kind of obscures how that works. In practice, we would have that `ImageSetInfo.type` is a number with limited allowed values (I think...), and it's going to be hard to remember what the various magic numbers mean when reading the raw JSON.

The way that enums are handled in other parts of this module (e.g. `classic_pywwt.CreateAnnotationMessage.shape`) is to use a TypeScript literal union type:

```
shape: "circle" | "line" | "polygon";
```

This makes the JSON representation much clearer to the reader, I think, and also results in more readable JSON data. It's a minor hassle to translate the TypeScript enum to this expression but I think it would be worth it here.

research-app-messages/src/selections.ts
```diff
@@ -8,9 +62,9 @@ export interface Source {
   ra: number;
   dec: number;
   name: string;
-  catalogName: string;
+  layer: LayerInfo;
   zoomDeg?: number;
-  catalogData: {
+  layerData: {
```
pkgw, line 67: Note that by removing the existing `catalogName` and `catalogData` fields, you are breaking the messaging API here. In this  case, I think that's OK since we have strong reason to believe that there is zero code that actually depends on the existing API so far. But in other cases it's important to preserve existing fields/behaviors/etc for compatibility.

Also note that the docstring will need updating since it references `catalogName`.

See also some comments on terminology that I'll have below.

research-app/src/App.vue
```diff
@@ -26,14 +26,14 @@
           v-bind:imageset="imageset"
         />
       </div>
-      <div id="catalogs-container" v-if="haveCatalogs">
+      <div id="spreadsheets-container" v-if="haveTableLayers">
         <div class="display-section-header">
-          <label>Catalogs</label>
+          <label>Catalogs and Spreadsheets</label>
```
pkgw, line 31: So, terminology. The "spreadsheet" term is pretty WWT-specific — I think people will understand it, but I don't see a need to go out of our way to surface it. People will understand "catalog" to mean either their own data table, or something like a HiPS progressive catalog. So I think the name of this section can stay as just "Catalogs". Or "Data Tables" might be slightly more generic.

Then we should be more limited in where we use the "layer" terminology since that can include not just spreadsheet layers (both "small" and HiPS progressive), but also imagery layers like FITS files. (And the WWT engine provides some more obscure layer types that we don't currently use in these UIs.)

research-app/src/App.vue
```diff
@@ -26,14 +26,14 @@
...
         </div>
-        <catalog-item
-          v-for="catalog of hipsCatalogs"
-          v-bind:key="catalog.name"
-          v-bind:catalog="catalog"
+        <spreadsheet-item
+          v-for="layer of layers"
```
pkgw, line 34: So, I the state item here should be named something like `tableLayers` instead of just `layers`. Looking farther down in the code, it looks like that's actually the name used in the Vuex store, and I definitely think we should tend to preserve those names to make things overall more consistent.

research-app/src/App.vue
```diff
@@ -312,7 +312,7 @@ import { mapGetters, mapMutations, mapState } from "vuex";
 
 import { distance, fmtDegLat, fmtDegLon, fmtHours } from "@wwtelescope/astro";
 
-import { Source, WWTResearchAppModule } from "./store";
+import { LayerInfo, Source, WWTResearchAppModule } from "./store";
```
pkgw, line 315: Here, I find the `LayerInfo` name confusing, since my default assumption would be that it would also include imageset layers as one of the enumeration options: "information about any kind of WWT layer". But this type actually has a more specific role. It's extra-confusing in this particular case since the type *does* include an ImagesetInfo option for HiPS progressive catalogs ... just not imageset *layer* info. I'd suggest renaming to something like `CatalogLayerInfo`, with a doc-comment at the type definition explaining why the Imageset option is there.

research-app/src/App.vue
```diff
@@ -2195,12 +2198,18 @@ export default class App extends WWTAwareComponent {
...
-      const layer = this.layerForHipsCatalog(catalogName);
-      if (layer == null) {
-        continue;
+    for (const layerInfo of this.visibleTableLayers()) {
+
+      let layer: SpreadSheetLayer | null = null;
+      if (isImageSetInfo(layerInfo)) {
+        layer = this.layerForHipsCatalog(layerInfo.name);
+      } else {
+        layer = this.spreadSheetLayerById(layerInfo.id);
```
pkgw, line 2207: *Maybe* it would be worth centralizing this logic somewhere? I feel like the code that handles the different enum-type cases should be found near the type ... if we, say, add a new option to the enum one day, we'd have to touch code in several different places to update it all to handle the new type.

research-app/src/SpreadsheetItem.vue
```diff
@@ -158,47 +169,64 @@ export default class CatalogItem extends Vue {
...
   isSelected = false;
+  isLayerHips = false;
+
+  layerId(): string {
+    return isImageSetInfo(this.layer) ? this.layer.name : this.layer.id;
+  }
+
+  layerState(): SpreadSheetLayerSettingsInterfaceRO | null {
+    const id = this.layerId();
+    return isImageSetInfo(this.layer) ? this.spreadsheetStateForHipsCatalog(id) : this.spreadsheetState(id);
```
pkgw, line 225: This is another area where maybe it's worthwhile to bring this logic closer to the definition of (what's currently named) the LayerInfo type.

pkgw (commented):
Good work! This all looks well-designed. I have some comments, but they're mostly about big-picture things. In particular I think that it would be good to try to reserve the "layer" terminology for cases that really mean *any* kind of WWT layer: tables, imagery, potentially other things too. This kind of decision gets exposed publicly in things like the source data structure relayed through the research app messages.

This PR has some conflicts against HEAD that will need to be resolved now, but I think they're pretty straightforward CSS/test-suite synchronization issues.

Carifio24:
I made changes to address your comments. Most of them were straightforward, but I think there are a few things worth noting:

- I renamed the list of spreadsheet layers in the app to `spreadsheetLayers`, as `tableLayers` is already being used for the map of table layer message handlers.
- I agree with your suggestion that we put the dispatch logic for the `CatalogLayerInfo` closer to the type definition. Since our use case it typically to grab something from the engine's Vuex store, I moved the definition of `CatalogLayerInfo` into the engine store (which is where its possible values already live), and put the relevant dispatch logic there. The various layer-grabbing functions were pretty redundant, so I created a `spreadSheetLayerByKey` function external to the store itself, and have the store functions call that as appropriate.

PR #147, opened 2021-09-24
Position fix for SkyImages when using WebGL 1

imbasimba:


pkgw:
Woohoo, great find!

pkgw:
Browserstack test just failed now due to a transient issue, but it worked before, so I'm going to go ahead and merge.

PR #148, opened 2021-09-27
Research app UI element sizing and positioning in response to screen size

Carifio24:
This PR modifies the research app UI to be more responsive to changes in screen size. This is done via flexbox layouts (sometimes nested). The main things that this layout looks to achieve are the following:

- Keep the selection tools centered on the screen while preventing any overlap with the display panel. The top-level flexbox containers that wrap the display panel, tools, and controls have `flex` values of 2, 3,  and 2, respectively, which means that the share of horizontal space allocated to the `display-panel-box` is 2/7, or about 28%. The display panel has a `max-width` of `25vw`.
  * Using 1,2,1 gives the display panel no extra space inside of its container (unless the browser is spanning multiple monitors), which I thought looked worse when shrinking the screen size. Using 4, 7, 4 (which gives a share of 4/15, or about 27%, to the display panel container), didn't look noticeably different than 2, 3, 2.
- Keep a minimum width (of `200px`) for the display panel, to preserve usability, until the screen width becomes very small.
- When the screen becomes narrow (less than `425px`), the selection tools wrap onto the next line, below the display panel and the controls, and are centered horizontally.
- For very narrow screens (narrower than `250px`), the display panel will shrink to accommodate the controls. I didn't want to have the controls wrap to the next line - since they're relatively tall and narrow, it doesn't look very good.

For the very narrow (sub-`250px`) screens some of the display panel items can get cut off. Once we get the large-scale behavior finalized I might go take another look at some of them, but I think there's also just a limit to what can be done with such a narrow screen.

### DIFF
--- a/research-app/src/App.vue
+++ b/research-app/src/App.vue
@@ -10,276 +10,278 @@
 
     <!-- keydown.stops here and below prevent any keynav presses from reaching
       the toplevel UI handlers -->
-    <div id="display-panel" v-if="!hideAllChrome" @keydown.stop>
-      <transition name="catalog-transition">
-        <div id="overlays">
-          <p>{{ coordText }}</p>
-        </div>
-      </transition>
-      <div id="imagery-container" v-if="haveImagery">
-        <div class="display-section-header">
-          <label>Imagery</label>
-        </div>
-        <imageset-item
-          v-for="imageset of activeImagesetLayerStates"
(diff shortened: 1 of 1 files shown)


pkgw (approved):
OK, I'm going to confess that I'm not reviewing this in nearly as much detail as PRs that touch the TypeScript, but from loading this up and playing with the window size, this looks great!

PR #149, opened 2021-10-09
Add PHAT imagery test and bolster test code

Carifio24:
This PR focuses on two items. First, I've added in a test of the WTML loading and imagery layers, using the [PHAT](http://data1.wwtassets.org/packages/2021/09_phat_fits/index.wtml) FITS data as the sample. This is a pretty straightforward test, similar to what we had for the HiPS catalogs plus the WTML loading - load the file, check that we have the right number of layers, then select the first layer and check that the display panel behaves correctly.

Second, I've tried to remove some of the obvious stress points that we'd noticed. In particular:
- I added a utility function to help with automatically creating `RegExp` objects from strings.
- I added some code to count the numbers of available HiPS catalogs and background imagesets from the appropriate URLs [here](https://web.wwtassets.org/engine/assets/builtin-image-sets.wtml) and [here](http://www.worldwidetelescope.org/wwtweb/catalog.aspx?W=hips). I assume these don't change often, but if they do, the tests won't break. Since we're using Node, I've added `axios` and `jsdom` to the dependencies to facilitate this operation.
- Most importantly, the `nth-child-of` selectors seem brittle, and like something that would require a lot of maintenance. After some digging around, I discovered that Nightwatch makes selecting particular elements easy to do (though not easy to find - I couldn't find this anywhere in the documentation, and just happened to stumble onto the relevant GitHub traffic). We can replace the selector with an object of the form `{selector: <selector>, index: <n>}`  to get the nth object that matches a query. This PR changes the existing tests to grab items using this method. I think this is a much better approach - most of our selectors are based on IDs or classes rather than element types, so if something we do breaks this, the test probably merits another look as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,7 @@
       "name": "wwt-webgl-engine-monorepo",
       "dependencies": {
         "@types/debounce": "^1.2.0",
+        "@types/jsdom": "^16.2.13",
         "@types/vue-color": "^2.4.3",
         "@types/vue-select": "^3.11.1",
         "@wwtelescope/astro": "file:astro",
@@ -19,7 +20,9 @@
         "@wwtelescope/engine-vuex": "file:engine-vuex",
         "@wwtelescope/research-app": "file:research-app",
         "@wwtelescope/research-app-messages": "file:research-app-messages",
+        "axios": "^0.22.0",
         "debounce": "^1.2.1",
+        "jsdom": "^15.2.1",
         "vue-color": "^2.8.1",
         "vue-select": "^3.11.2"
       },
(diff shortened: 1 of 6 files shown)


pkgw:
Maybe try testing with a fresh checkout? If all of the different browsers are failing in the same way but you can't reproduce locally, I feel like the CI system must somehow be running different code than you are.

Carifio24:
That was a good idea...but even off a fresh clone of the branch, I still can't manage to reproduce this error locally.

pkgw:
I've rebased this branch onto the latest `master` to pull in the fix for (avoiding) Node 17.

PR #155, opened 2021-10-29
research-app: Support for initializing app state via query parameter

Carifio24:
This PR adds support for the research app to initialize its state via a query parameter. In particular, this implementation uses a `messages` query parameter. The expected content of this parameter is a base64 encoding of a list of messages. Additionally, this PR adds a tool which gives the user a URL which can reproduce their current state (with the exception of user-added table layers).

Since some messages are intended to modify already-existing objects (e.g. changing settings on a table layer), we need to execute messages in the correct order to accurately reproduce the state. If all of our messages were synchronous, we could just do a topological sort of our messages (based on message type), and execute them in order. That's not the case here - some of our messages are asynchronous. Thus, this code does more of a topological 'spreading' through the space of messages, starting with a set of 'root' messages that don't rely on anything else in the list, with dependent messages added to the message queue only when their prerequisites are met. For synchronous messages, as mentioned above, we just make sure that any dependent messages fall after them in the queue. However, for asynchronous messages, we attach a message handler for the reply message type. When all of the messages of the type have received their reply, we delete the handler and can move on.

Having to listen for these reply messages means that using this initialization also requires setting the `origin` query parameter to the app itself, so that it can listen to its own messages. I think this is necessary in order to have this work within the message framework.

In order to facilitate exporting the app state to a URL, I added three messages. I placed the last two messages in `layers.ts`, even though they feel more akin to what's in `classic_pywwt.ts`, since they weren't created for older version of `pywwt`.
* `AddSourceMessage` is straightforward - it adds a given source to the app. If a layer with the same name as the source exists, this layer is found and set as the source's `catalogLayer`. Otherwise, a new `CatalogLayerInfo` object is created.
* `MultiModifyTableLayerMessage` allows modification of multiple table layer settings in one message. This prevents the need to have a large list of settings messages for each layer - as we don't know which settings have changed from their default values, we need to send every setting along.
* 'MultiModifyFitsLayerMessage` does the same thing for imageset layers.

I also made a small modification in `engine-types`  to help with parsing of `Date` objects.


research-app-messages/src/selections.ts
```diff
@@ -85,17 +85,32 @@ export interface Source {
...
+  /** The tag identifying this message type */
+  type: "add_source";
+
+  /** The source to be added */
+  source: Source;
+}
+
+export function isAddSourceMessage(o: any): o is AddSourceMessage {
+  return typeof o.type === 'string' &&
+      o.type == 'add_source' &&
```
pkgw, line 113: This could just be a triple-equals `o.type === add_source`

research-app/src/App.vue
```diff
@@ -116,6 +116,17 @@
...
+                  <li>
+                    <a
+                      href="#"
+                      v-close-popover
+                      @click="
+                        selectTool('save-state');
+                        showPopover = false;
+                      "
+                      tabindex="0"
+                    ><font-awesome-icon icon="save" /> Save current state</a>
```
pkgw, line 128: I like where you're coming from here but I think the average user might not "get" why the URL export is the same as saving the state. I think a more intuitive title for this would be "Create link to current view", or maybe "Create link to current state".

research-app/src/App.vue
```diff
@@ -286,12 +297,32 @@
                 </div>
               </div>
             </template>
+            
+            <template v-else-if="currentTool == 'save-state'">
+              <div class="save-state-container">
+                <label class="save-state-title">The current app state can be restored using:</label>
```
pkgw, line 303: The descriptive text here should track the change in the tool description, above.

research-app/src/App.vue
```diff
@@ -365,6 +405,15 @@ import {
   convertPywwtSpreadSheetLayerSetting,
   convertSpreadSheetLayerSetting,
 } from "./settings";
+import { imageSetLayerSettingNames, ImageSetLayerState } from "@wwtelescope/engine-helpers/src/imagesetlayer";
+import { isSpreadSheetLayerSetting, spreadSheetLayerSettingNames, SpreadSheetLayerState } from "@wwtelescope/engine-helpers/src/spreadsheetlayer";
+import { isLoadImageCollectionCompletedMessage, isLoadImageCollectionMessage, PywwtSpreadSheetLayerSetting } from "@wwtelescope/research-app-messages/dist/classic_pywwt";
+import { type } from "os";
```
pkgw, line 411: Is this `os.type` import used anywhere? It's hard to search for, but I couldn't see a usage? And it looks like this may be a special Node.js module — is it one of these things that TypeScript specially handles?

research-app/src/App.vue
```diff
@@ -791,6 +905,22 @@ class AnnotationMessageHandler {
     }
   }
 
+  handleMultiModifyAnnotationMessage(msg: layers.MultiModifyAnnotationMessage) {
+    for (const [index, option] of msg.settings.entries()) {
+        const setting: [string, any] = [option, msg.values[index]];
+        if (this.ann instanceof Circle && isCircleAnnotationSetting(setting)) {
```
pkgw, line 911: Indentation got a little wonky here.

research-app/src/App.vue
```diff
@@ -1213,6 +1361,369 @@ export default class App extends WWTAwareComponent {
...
+
+  decodeObjectBase64(data: string): object {
+    return JSON.parse(Buffer.from(data, 'base64').toString());
+  }
+
+  adjustSettingsForImport(names: string[], values: any[]): void {
+    for (let i = 0; i < names.length; i++) {
+      const name = names[i];
+      const value = values[i];
+      if (name === 'color') {
```
pkgw, line 1384: I don't *love* this special-casing ... this sort of specialization and de-specialization is the sort of thing that the "pywwt" types in research-app-messages handle. Is there a way to leverage that existing infrastructure?

research-app/src/App.vue
```diff
@@ -1213,6 +1361,369 @@ export default class App extends WWTAwareComponent {
...
+  }
+
+  stateAsUrl(): string {
+
+    const coordinatesMessage: classicPywwt.CenterOnCoordinatesMessage = {
+      event: "center_on_coordinates",
+      ra: this.wwtRARad * R2D,
+      dec: this.wwtDecRad * R2D,
+      fov: this.wwtZoomDeg / 6,
+      instant: true,
```
pkgw, line 1545: This piece is failing to preserve the rotation of the view. You can demonstrate this by holding down Control and click-dragging to rotate the camera along the "roll" axis.

research-app/src/App.vue
```diff
@@ -1213,6 +1361,369 @@ export default class App extends WWTAwareComponent {
...
+        loadCatalogsMessages.push({
+          event: "layer_hipscat_load",
+          threadId: threadId, // We need this to get a response
+          tableId: id,
+          name: catalog.name,
+        });
+
+
+        const state: SpreadSheetLayerState = this.wwtSpreadSheetLayers[id];
+        const values: any[] = spreadSheetLayerSettingNames.map(setting => (state as any)["get_" + setting]());
```
pkgw, line 1586: This kind of functionality is centralized in functions like [`@wwtelescope/engine-helpers::extractSpreadSheetLayerSettings`][x]. Can they be used here?

[x]: https://docs.worldwidetelescope.org/webgl-reference/latest/apiref/engine-helpers/modules.html#extractspreadsheetlayersettings

research-app/src/App.vue
```diff
@@ -1213,6 +1361,369 @@ export default class App extends WWTAwareComponent {
...
+      ...sourceMessages,
+      ...createAnnotationMessages,
+      ...annotationSettingsMessages,
+    ].flatMap(s => s ? [this.encodeObjectBase64(s)] : []);
+
+    const messageString = messageStrings.join(",");
+    const outString = messageString;
+
+    const params = {
+      origin: window.location.origin,
```
pkgw, line 1706: Hmmm. I think the best thing here is not the specify the `origin` query-string parameter? If someone's going to be pulling open this link later, it'll open up a freestanding version of the research app, and I think it makes the most sense for that not to allow `Window.postMessage` calls from anywhere.

research-app/src/App.vue
```diff
@@ -1213,6 +1361,369 @@ export default class App extends WWTAwareComponent {
...
+      messages: outString,
+    };
+
+    return window.location.origin + '/?' + (new URLSearchParams(params)).toString();
```
pkgw, line 1710: This isn't generically correct since `window.location.origin` strips off the URL path of wherever the app instance is being hosted. I think the proper way to do this is copy `window.location` as a parsed-out URL, overwrite its query string, and then stringify the resulting URL.

pkgw (commented):
I have to admit that the message sorting code feels very complex to me, but dude, the result is definitely *extremely* cool. Being able to "save" the app state and restore it is very awesome, and I can see how there needs to be some dependency analysis at some point in the process.

I noticed a few things to fix, and there's so much new code here that I think there will probably be some further things to shake down. But I don't think it's going to be possible to add this much functionality with less code, so I think we'll have to go ahead, get the feature in, and tackle any issues as they arise.

I've updated your branch to merge in `master` to solve the lockfile conflicts.

research-app/src/App.vue
Carifio24: Yeah, that's a good point. I'm thinking the alternative is, if the `messages` parameter is present, to add the current origin as the allowed origin in `main.ts` (that is, on startup, rather than relying on a query parameter from when the state was serialized).

research-app/src/App.vue
Carifio24: Is there currently a way to access the "roll" value of the view in the app?

If not, it seems like it would be pretty straightforward to add - I'm guessing it's related to one, or both, or `angle` and `rotation` in the engine's `CameraParameters`.

research-app/src/App.vue
Carifio24: Hmm, this isn't used anywhere (in fact, VS Code has it marked as unused). This feels like a mistaken auto-import to me (e.g. I typed, `os.type` instead of `o.type` for a message or something), and fixed the typo but not the import.

research-app/src/App.vue
Carifio24: I was able to remove the special handling for `color` using the existing pywwt message conversion, but (unless I'm missing something) I don't see a way to do the same for the `Date`s. The issue is that the dates are being serialized as strings, but both spreadsheet layer setting types are expecting `Date` objects.

Depending on how we want users to pass in dates in messages (is a string fine, or do we really want a `Date` object), this code could be moved into the handler's `handleModifyMessage` method.

research-app/src/App.vue
Carifio24: I did what you suggested here - I created a `URL` object from the window location, overwrote its `search` member, and obtained the string form of the new URL.

Carifio24:
Most of my updates were pretty straightforward from your suggestions, but I just wanted to note a few:
* In order to allow preserving the camera's roll, I exposed the `rotation` value from the store's `renderContext.viewCamera` to the `WWTAwareComponent` as `wwtRollRad`.
* Like we talked about, I was able to remove the `Color` special-casing, but not the `Date`. I left a comment explaining the reason for that.
* Instead of giving the origin as a query parameter, I now have `main.ts` check if there are any messages in the URL query string. If there are, it sets the current origin as the allowed origin so it can listen to its own messages.

PR #160, opened 2021-11-23
Filled annotation circle performance improvement

imbasimba:
This reduces the memory usage of filled circles by >98%. While this can be improved further, this is the 20% of the effort giving 80% of the value. 

The performance is still not optimal when showing thousands of filled circles, but now it is at least functional. 

With the previous implementation the inside of the circle was filled with 1136 triangles. Each of these triangles have 3 points. Each of the points have a position (x, y, z), a normal (x, y z), a start date and end date, a color (rgba). So, in total, filling a circle required 1136 * 3 * (3 + 3 + 1 + 1 + 4) = 40 896 numbers. So no wonder there were some issues :smile:

The new implementation reduces the number of points to 74 and removes the normal -> 74 * (3 + 1 + 1 + 4) = 666 numbers per filled circle. 

With some changes, we could reduce it to 228 numbers, since we only really need one set of the dates and color. But this would require breaking some current engine conventions to implement. 

In theory, I think it should be possible to draw a quad (two triangles), and use the fragment shader to only fill within a radius from the quad center, which would draw a circle. This, together with the above improvement would draw the same shape with 4 * 3 + 1 + 1 + 4 = 18 numbers. So there is another ~98% memory reduction available should we need it.

The utopian solution (using only 9 numbers) of creating the circle in the fragment shader with a single position as input due to limitation in point size (ALIASED_POINT_SIZE, which is also different for different systems). This limitation causes the circle to not grow when zooming past a certain point.

### DIFF
--- a/engine/wwtlib/Annotation.cs
+++ b/engine/wwtlib/Annotation.cs
@@ -15,6 +15,7 @@ namespace wwtlib
 
         protected static PointList PointList = null;
         protected static LineList LineList = null;
+        protected static TriangleFanList TriangleFanPointList = null;
         protected static TriangleList TriangleList = null;
 
         public static bool BatchDirty = true;
@@ -24,6 +25,7 @@ namespace wwtlib
             {
                 PointList = new PointList(renderContext);
                 LineList = new LineList();
+                TriangleFanPointList = new TriangleFanList();
                 TriangleList = new TriangleList();
                 LineList.DepthBuffered = false;
                 TriangleList.DepthBuffered = false;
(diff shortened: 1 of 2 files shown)


astrojonathan:
I have a version of this code that will run on two triangles and use a shader to fill the Circles.  We might consider doing a perf comparison. The four points insure the circle will grow and shrink with zoom.

imbasimba:
Cool! I'd love to see how you implemented that! I would be surprised if the two triangles strategy isn't performing noticeably better.

pkgw:
I certainly wouldn't mind getting that extra factor of 50, but this is very cool — thanks Henrik!

Issue #34, opened 2019-01-11
Issue with alignment of Milky Way


astrofrog:
Using Astropy, I generated a bunch of Galactic coordinates with a range of l and with b=0. I then converted these to RA/Dec and plotted them at a distance of 10kpc:

```js
wwt.setBackgroundImageByName('3D Solar System')
csv = 'ra,dec,distance\r\n86.40498828654476,28.936177761791473,10000.0\r\n92.76903277368984,18.87863032323945,9999.999999999998\r\n98.39984402308887,8.629663181475806,10000.0\r\n103.72919505205704,-1.6969502306914939,10000.000000000004\r\n109.1162686148206,-12.008742481391023,10000.000000000002\r\n114.93391443514108,-22.209283594632566,10000.000000000002\r\n121.66028106252394,-32.17084852335976,10000.0\r\n130.0098814512841,-41.69016314011403,10000.0\r\n141.1284792791373,-50.40144173479356,10000.0\r\n156.72234785174578,-57.60886784036625,9999.999999999998\r\n178.19841226529226,-62.09467439477692,10000.0\r\n203.3153363453848,-62.480512337639325,10000.0\r\n225.73834237457388,-58.61392886680624,10000.0\r\n242.2863954395057,-51.77171215352405,10000.0\r\n254.0382842268111,-43.25644805982758,9999.999999999998\r\n262.7662689413582,-33.843310483992944,10000.0\r\n269.7112608376061,-23.940124158307032,10000.0\r\n275.64628922937965,-13.770128469571242,10000.0\r\n281.0791151279341,-3.470099332026524,10000.0\r\n286.39350207725835,6.860352794351721,9999.999999999998\r\n291.9451087925097,17.129778580923652,10000.0\r\n298.1463542847858,27.230148032147596,10000.0\r\n305.57333894325194,37.00336126063327,10000.0\r\n315.12862349903395,46.17991830838576,10000.0\r\n328.24524299089103,54.25591541337054,10000.0\r\n346.7277117354867,60.287407155918324,10000.0\r\n10.675060504980387,62.85484124259376,10000.0\r\n35.205682940649886,61.01583466693159,9999.999999999998\r\n54.728803394410946,55.4711940622356,10000.0\r\n68.64163336213086,47.66338198628247,10000.000000000002\r\n78.68913167687124,38.6306333324857,9999.999999999998\r\n86.40498828654476,28.936177761791473,10000.0\r\n';
[...]

PR #166, opened 2022-01-06
Fix SpreadSheetLayer longitude values

Carifio24:
This PR is aimed at fixing the cause of #34. I think the culprit is [SpreadSheetLayer.GeoTo3dDoubleRad](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Layers/SpreadSheetLayer.cs#L826), which adjusts the longitude of the point by 180 degress before performing the transformation.

This function is currently called in five places:
[SpreadSheetLayer.cs, line 826](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Layers/SpreadSheetLayer.cs#L826)
[SpreadSheetLayer.cs, line 1211](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Layers/SpreadSheetLayer.cs#L1211)
[SpreadSheetLayer.cs, line 1212](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Layers/SpreadSheetLayer.cs#L1212)
[SpreadSheetLayer.cs, line 1270](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Layers/SpreadSheetLayer.cs#L1270)
[Coordinates.cs, line 111](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Coordinates.cs#L111)

In all five of these cases, the Windows client calls `SpreadSheetLayer.GeoTo3dDouble` - specifically, the one with the three-argument signature, which performs the same operation, minus the longitudinal adjustment. The corresponding Windows client calls are [here](https://github.com/Carifio24/wwt-windows-client/blob/master/WWTExplorer3d/SpreadSheetLayer.cs#L1332), [here](https://github.com/Carifio24/wwt-windows-client/blob/master/WWTExplorer3d/SpreadSheetLayer.cs#L1790), [here](https://github.com/Carifio24/wwt-windows-client/blob/master/WWTExplorer3d/SpreadSheetLayer.cs#L1791), [here](https://github.com/Carifio24/wwt-windows-client/blob/master/WWTExplorer3d/SpreadSheetLayer.cs#L1848), and [here](https://github.com/Carifio24/wwt-windows-client/blob/master/WWTCore/WWTCore/Coordinates.cs#L130).

The web engine function `GeoTo3dRad` (defined [here](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Coordinates.cs#L32)) corresponds to `GeoTo3dDouble` in the Windows client. The names are a bit wonky here - maybe an attempt to avoid overloads for ScriptSharp?). So, for this PR, I've changed all of the calls to `GeoTo3dDoubleRad` to `GeoTo3dRad`.

### DIFF
--- a/engine/wwtlib/Coordinates.cs
+++ b/engine/wwtlib/Coordinates.cs
@@ -108,7 +108,7 @@ namespace wwtlib
                     meridean = -meridean;
                 }
             }
-            return Coordinates.GeoTo3dDoubleRad(lat, 90 + lng + meridean, radius);
+            return Coordinates.GeoTo3dRad(lat, 90 + lng + meridean, radius);
         }
 
         //static public Coordinates EquitorialToHorizon4(Coordinates equitorial, Coordinates location, Date utc)
--- a/engine/wwtlib/Coordinates.cs
+++ b/engine/wwtlib/Coordinates.cs
@@ -65,7 +65,7 @@ namespace wwtlib
             x -= falseEasting;
             y -= falseNorthing;
 
-           
+
 
             if (x != 0 || y != 0)
             {
@@ -280,7 +280,7 @@ namespace wwtlib
 
             return Vector2d.Create(hrAngle,dec);
         }
-      
+
         //static void AltAzToRaDec2(double alt, double az, out double hrAngle, out double dec, double lat)
         //{
         //    if (alt == 0)
@@ -290,7 +290,7 @@ namespace wwtlib
         //    if (az == 0)
         //    {
         //        az = .00000000001;
-        //    } 
+        //    }
         //    double sin_dec;
         //    double cos_lat = Math.Cos(lat);
 
@@ -400,7 +400,7 @@ namespace wwtlib
             return mst;
         }
 
-        
+
 
         public double RA
         {
@@ -431,7 +431,7 @@ namespace wwtlib
         {
             get
             {
-                
+
                 return declination / RC;
             }
             set
@@ -457,7 +457,7 @@ namespace wwtlib
 
             }
             //todo This was broken check callers to see what effect it had.
-            set 
+            set
             {
                 ascention = ((value*RC)+(Math.PI*2)%(Math.PI*2));
             }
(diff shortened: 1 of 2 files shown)


pkgw:
Hmm. I believe that we need to do one extra thing here, because this change introduces an incorrect 180° flip for table layers viewed in the 2D sky mode. I checked this by loading up the NASA Exoplanet Catalog demo notebook in [pywwt-notebooks] using the patched code, where an early example should focus in on the Kepler field with a dense grid of exoplanets.

[pywwt-notebooks]: https://github.com/WorldWideTelescope/pywwt-notebooks/

This must be related to @astrofrog's patch from a while ago:

https://github.com/WorldWideTelescope/wwt-webgl-engine/commit/8bccebf508637eea2637ddf775760dcdfff35edc

Note that `bufferIsFlat` is a flag indicating whether we're in a 2D sky-like mode, as opposed to a 3D mode. My guess is that the port to ScriptSharp introduced some issue that confused the 180° flips depending on `bufferIsFlat`, and that these were compensated for by (1) the 180° offset in GeoTo3dDoubleRad and (2) the 180 offset here:

https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/a6ec3e40794035f2ae89213f0a6bebed20ba1d80/engine/wwtlib/Layers/SpreadSheetLayer.cs#L814

which *is* present in the Windows client, being commented out, which Tom then un-commented to fix 2D table layers.

To get things going quickly, I think the right move is to re-comment line 814, which I think should yield correct behavior for both 2D and 3D views. This means that the web engine still has some kind of 180 inversion relative to the Windows engine, which I think has to stem from some issue related to ScriptSharp.

I think we should also remove  the 180° offset from Coordinates.cs, or delete the function in question entirely. It's just really really broken for that function to have that weird flip secretly inserted. I'm sure it was added to compensate for some flip that was showing up in the webclient, but that's just not the right way to make the problem go away.

pkgw:
OK, I've implemented the above and I believe that things are behaving as expected. Longer term it would be really good to figure out what "flipped" in the web engine to trigger this cascading set of hacks.

Closes #34.

PR #168, opened 2022-01-23
Fixed goto position for tiled images with offset and SkyImages rescaled to 256x256

imbasimba:
Some FITS files like the below one did not pan to the correct location.
[FITS file](https://github.com/WorldWideTelescope/wwt-webgl-engine/files/7920180/zoom-pos.zip)


imbasimba:
I think the pan position looks more correct with this fix than without it for the PHAT imageset that is failing in browserstack. So I still think this fix is correct and that the test needs to be updated. I welcome another pair of eyes to have a look at this though, to make sure I haven't made any mental blunders.

pkgw:
I've pushed a merge of `master` to get the fix for the CI.

pkgw:
Closing/reopening to retrigger the build here.

PR #169, opened 2022-01-30
Vibrations fix

imbasimba:
Using two 32-bit numbers to describe position of tiles to fix vibrations for very small (~arcseconds and less) tiles. So instead of having a single number pin point the vertex location, we have two. One to set the general area, and one to set the relative position from the first position. 

Here is a FITS file that vibrated a lot before this fix:
[vibrations.fits.gz](https://github.com/WorldWideTelescope/wwt-webgl-engine/files/7965582/vibrations.fits.gz)


imbasimba:
I just found an issue with this PR, so I converted it to a draft PR until I've got it all sorted out

pkgw:
I've pushed a merge of `master` to get the fix for the CI.

pkgw (approved):
OK, I sat down to give this a test — looking good! It looks like BrowserStack is having an intermittent problem here but I don't believe that it's related to this PR.

Issue #53, opened 2020-07-15
engine: missing dep on / reference to `pako` library


pkgw:
`engine/wwtlib/Layers/Layer.cs` includes a literal reference to the `pako` decompression library. We don't include that code right now, so the engine crashes in certain usages.

Triggered by https://bit.ly/wwt-neowise2020 , which loads http://data1.wwtassets.org/packages/2020/07_neowise/Comet_NEOWISE_July_2020.wtt .

Carifio24:
I've looked into this a bit (since it should be relevant for tour export in glue). We had talked before about doing something like `const pako = import('pako')` to grab the `pako` library. The one issue that makes that a bit thorny is that when `import` is used dynamically like that, it returns a promise rather than the library object itself.

That being said, we can use import if we replace [this line](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Layers/Layer.cs#L647) with something like

```
  Script.Literal("if (typeof pako == 'undefined') { import('pako').then((pako) => {");
  result = (string)Script.Literal("pako.inflate(e.target.result, { to: 'string' })");
  Script.Literal("}); } else { ");
  result = (string)Script.Literal("pako.inflate(e.target.result, { to: 'string' })}");
```

Up to some formatting, that makes the output JavaScript in that block be
```
try {
    if (typeof pako == 'undefined') {
      import('pako').then((pako) => {;
        result = pako.inflate(e.target.result, { to: 'string' });
      });
    } else { ;
      result = pako.inflate(e.target.result, { to: 'string' })
    };
  }
```

Not the most elegant fix, but it would work.

Worth noting that in a CommonJS setting (e.g. Node), we could use `require`, which is synchronous.

Carifio24:
Ah, yeah, as written this will almost certainly not work. It is good to know that we can get ScriptSharp to generate promise-y code with the assignment to `result` still being in C# (which works since the extraneous semicolons are just ignored).

Generating stub classes to translate to JS sounds interesting; I'll look into that.

pkgw:
Cool. Stubs classes won't help with the core issue of async imports here, but again, I'm sure there's a way to solve that one.

PR #170, opened 2022-02-08
Import pako in static constructor block

Carifio24:
This PR provides a fix for #53 by adding a piece of code to import the `pako` library and assign it to a variable of the same name. To do this, I created a static `Imports` class and put the relevant `Script.Literal` statements into the class's static constructor. It would be nice to have something more robust in the future, but this promise should resolve before pako is needed in pretty much every case (as an example, the embed link referenced in the issue works fine with this fix).

One thing I'm a bit unsure about is if we want to add `pako` to `package.json` (either at the top level or the engine level), since we're explicitly referencing it. It's already in the top-level `package-lock.json` (line 38324). It's a dependency of `browserify-zlib`, which is a dependency of `node-libs-browser`, which is a dependency of `webpack`, which is a dependency of some of the `vue-cli` packages.


pkgw (commented):
As per our conversation, yeah, we should add pako as an explicit dep.

For what it's worth, I think that the arrow function here is a more modern JavaScript feature than this codebase has required up to this point. But looking at [CanIUse][ciu], it's super well established, and we probably require more modern browsers anyway for all of our WebGL stuff. So I'm not worried about this being a compatibility issue in practice.

[ciu]: https://caniuse.com/arrow-functions

Carifio24:
I'm also not sure what's up with the checks. Per the logs, the error is with adding an icon to the FontAwesome library in the `embed-creator`. That obviously didn't change at all here, and the type error that it's giving just seems wrong (`IconDefinitionOrPack` is defined as `IconDefinition | IconPack`).

pkgw:
I got an email about FontAwesome 6 coming out recently, so my guess is that we're pulling in a new version of that dep and it's breaking something. We might need to restrict the version that we use for a little while if there's some breakage in a new release ... or maybe we're using something in a wonky way that's no longer allowed.

Carifio24:
FontAwesome 6 was released 10 days ago, so I suspect you're right and that's the culprit. I'll see if I can get this working with some more specific pins, and then look into what we need to do to move on to FA6.

Carifio24:
Pinning `@fortawesome/fontawesome-svg-core` to `~1.2.28` seems to have done the trick as far as FontAwesome goes. I suspect that the `^1.2.28` pin was picking up `1.3.0`, which was released 8 days ago and likely is expecting FA6.

There's now an issue with the actual engine tests and `pako` that I'm not seeing locally, so I'll see what's up with that.

Carifio24:
So I think what's happening is basically what we were thinking won't happen in the wild - in the test, the `import` promise hasn't resolved before the tests run. This wasn't an issue in the test before because `pako` is grabbed there via a [script tag](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/tests/tests.html#L7). After hacking with `engine/src/index.js` a bit, it looks like `pako` defined inside `$exports` is shadowing the window-level `pako`.

One way around this is could be to check `typeof window !== "undefined" && "pako" in window` before importing pako (for scoping purposes, we'd then define the pako in `$exports` to be the global value). This will avoid the import if we're in a browser environment where we've already imported pako from a CDN, but will trigger the import otherwise. It feels hack-y to me, but it should work.

Carifio24:
I've implemented the workaround, and added a comment explaining why we can't create a helper function to do this given the library name (it looks like we'd need C# 10 in order to pull that off). While I was at it, I changed the arrow function to a regular anonymous function.

The BrowserStack test failed the first time through. Based on the logs it looks like the connection between BrowserStack and the CI app got disrupted, and everything was fine the second time through.

PR #173, opened 2022-02-16
Fix package.json typos

Carifio24:
While updating the pins for `fontawesome-svg-core`, I noticed that multiple of the package-level `package.json` files have a typo in the repository URL. This PR fixes those typos.


pkgw:
I merged master in to pick up the versioning fix, and also tried to update the package-lock to pick up all of the new package versions that dependabot is filing pull requests for.

PR #178, opened 2022-03-21
Use v5 UUIDs for catalog HiPS layer IDs

Carifio24:
Currently, the internal engine ID for HiPS catalog layers is simply the name of the layer's dataset. This is different behavior than every other layer, which use a UUID as the layer's ID. Aside from just not being a great practice, this is fundamentally incompatible with the Windows client. The Windows client uses C#'s [System.Guid](https://docs.microsoft.com/en-us/dotnet/api/system.guid?view=net-6.0) to represent layer ID (rather than the hand-rolled implementation here), which won't accept an ID that doesn't have the proper format.

To solve this issue, this PR modifies the engine so that HiPS layers are stored with a proper UUID. The approach here uses v5 UUIDs created from the layer's dataset name. This is something of a middle ground between a random v4 UUID and the current system - the UUID created from a given string will always be the same, and will have a standardized format that the Windows client will accept. The creation of a v5 UUID requires a "namespace", which is itself a UUID. Here I've used a UUID that I created via the `uuid` package's CLI tool. Note that this relationship only goes one way - even knowing the namespace, we can't recover the dataset name from the UUID. Generation of the v5 UUIDs is done using the `uuid` package, which is imported using the same mechanism that we use for `pako`.

There are a few changes to the Vuex pieces of the engine to account for this. This primarily consists of adding an `id` field to `ImagesetInfo` which is non-null if the imageset has an associated HiPS catalog, and fixing the functions that map layers to the key to use in the engine. I kept the various key-mapping layers of indirection for potential flexibility in the future. I also took the opportunity to simplify the mechanics of the HiPS catalog chooser in the research app - we didn't really need `catalogToAdd`; an input listener on the selection element is enough.

Finally, I made a change to how these layers are exported to (and imported from) tours. For catalog HiPS layers, rather than writing out the `SpreadSheetLayer`, instead an `ImageSetLayer` with the same name and ID is written out. This serves two purposes:
* The Windows client handles HiPS differently than the web engine - there, the HiPS layer is an `ImageSetLayer` with an associated VO table. While I have some slight updates to the Windows client that will be necessary, doing this makes the process pretty painless.
* Loading in the `ImageSet` in the `ImageSetLayer` makes it each for the web engine to load these layers as well - if the XML layer is a catalog HiPS layer (defined as having the Healpix projection type and `.tsv` extension), we just call `addCatalogHips` using the imageset, rather than loading the layer as normal.


Carifio24:
It looks like the BrowserStack test is failing because there's something strange happening with the imagery and background dropdowns. Not sure what the issue is yet, but I can't imagine it's anything too major.

pkgw (approved):
Looks great! No suggestions for the implementation. And hopefully BrowserStack will be easy to straighten out.

Carifio24:
So I cleaned up a few more name-based things that I had missed before. However, there is something funky going on with the dropdowns. I'm not sure whether it's something from this PR or a change in `vue-select`, but both the imagery and background dropdowns aren't working - if I try to make a change, nothing happens (the catalog dropdown, for whatever reason, is fine). I'm still working on getting to the bottom of it.

Carifio24:
Welp, I figured out the issue with the dropdowns. `vue-select` will, by default, JSONify an object...[unless there's an `id` field present](https://vue-select.org/api/props.html#getoptionkey), which just happens to be what I added to `ImagesetInfo`, and is always `null` for non-HiPS catalogs. The fact that every ID was null was breaking these dropdowns' ability to distinguish between items. The catalog dropdown was fine because the catalog layers all have a unique ID.

To fix this, I set the `getOptionKey` prop to be `JSON.stringify` for the background and imagery dropdowns, to match the default `vue-select` behavior.

pkgw:
Well that's a random surprise! Good work chasing that down.

PR #181, opened 2022-03-23
TOAST FITS support

imbasimba:
Moved FITS processing to base class (Tile.cs). Now any tile (including TOAST) using the base class implementation of MakeTexture and RequestImage is able to reder FITS images.

Bonus points for having more row deletions than additions :)


pkgw:
Bonus points indeed!

PR #183, opened 2022-03-25
Allow toggling whether sources from a given layer are selectable

Carifio24:
This PR adds the ability to toggle whether or not the sources in a given table layer are selectable. This toggling can be done either in the UI or programmatically via a message. My motivation for this was from investigating the `glue-wwt` speed issue when using the newest version of `pywwt`. I believe that issue pops up when there are a lot of points visible due to the nearest-source calculation slowing things down. Removing layer selectability should increase performance - in particular, if no layers are selectable, `closestInView` doesn't need to do any work. From testing on my machine, this did indeed seem to be the case. Aside from glue, though, I think any client embedding the research app benefits from this if they don't want source selection - performance should be increased and they won't see unnecessary cursor changes as the mouse moves off of/onto sources.

To accomplish this, an additional `selectable` field is added to the `TableLayerStatus` items in the research app's store. The closest-source algorithm in the research app then only looks for sources in layers that are both visible and selectable. Two messages are added - one to set the selectability of a particular layer, and another convenience message type to allow modifying the selectability of all layers at once. The single-layer message is handled by the `TableLayerMessageHandler` class in the research app. Since a user won't know the internal name of layer they create, we need to identify the layer by the name they gave it, which these handlers are already set up to do.

In the UI, this is accomplished by adding an extra icon in `SpreadsheetItem` which toggles selectability. I couldn't find a good FontAwesome icon that I felt made the intent obvious, so I "made" one by stacking the mouse pointer icon with the slash. Suggestions of something different to use here are welcome.

### DIFF
--- a/research-app-messages/src/selections.ts
+++ b/research-app-messages/src/selections.ts
@@ -112,3 +112,37 @@ export function isAddSourceMessage(o: any): o is AddSourceMessage {  // eslint-d
   return o.type === 'add_source' &&
       typeof o.source === 'object';
 }
+
+export interface ModifySelectabilityMessage {
+
+  /** The tag identifying this message type */
+  type: "modify_selectability";
+
+  /** The identifier of the layer to modify. */
+  id: string;
+
+  /** Whether to make the layer selectable. */
+  selectable: boolean;
+}
+
+/** A type-guard function for [[ModifySelectabilityMessage]]. */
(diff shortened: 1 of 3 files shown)


Carifio24:
Per our conversation this morning, I've removed the UI part of this PR.

PR #190, opened 2022-04-08
research-app: Allow creation of tour with current display

Carifio24:
This PR adds functionality for the research app to export its current view as a one-slide tour. In the Vuex layer, this is done via the `viewAsTourXml` action. The implementation of this is adapted from the `glue-wwt` [tour-saving code](https://github.com/glue-viz/glue-wwt/blob/main/glue_wwt/viewer/tools.py#L35) that worked in older versions of `pywwt`. That tool is the obvious use case for this, but maybe it will be helpful for other users as well.

From the client side, this is accomplished via a `GetViewAsTourMessage`, which promps the app to emit a `GetViewAsTourReply`. I didn't add a way to access this functionality in the UI. The one thing to note here is that the bundling of the tour as XML into the reply message is very straightforward, and there's the potential for the emitted message to be rather large if the current view contains a lot of data. For now, though, this gives a basic implementation of this functionality and will allow this to work in `glue-wwt` with future `pywwt` versions.


engine/src/index.d.ts
```diff
@@ -1961,8 +1961,20 @@ export class TourDocument {
...
   getTourStopIndexByID(id: string): number;
+
+  /** Get the tour document as an XML string */
+  getTourXML(): string;
+
+  /** Get a representation of the tour as a blob */
+  saveToBlob(): Blob;
 }
 
+export class TourEditTab {
```
pkgw, line 1972: It would be good to have some basic doc-comments for this new export and its methods.

research-app-messages/src/tours.ts
```diff
@@ -0,0 +1,32 @@
+// Copyright 2021 the .NET Foundation
```
pkgw, line 1: Copyright year 2022.

research-app/src/App.vue
```diff
@@ -759,6 +756,7 @@ class TableLayerMessageHandler {
     if (msg.settings.length !== msg.values.length) return;
 
     const layer = this.owner.spreadSheetLayerById(msg.id);
+    console.log(layer);
```
pkgw, line 759: I'm guessing this is some debugging code from unrelated work? It shouldn't be included here. (I find `git add -p` to be helpful for this kind of stuff.)

research-app/src/App.vue
```diff
@@ -780,6 +778,7 @@ class TableLayerMessageHandler {
           id: this.internalId,
           settings: layerSettings,
         });
+        console.log(JSON.stringify(layer));
```
pkgw, line 781: Same as previous comment.

pkgw (changes requested):
I love how straightforward this pull request is! A couple of very minor things to tidy up but other than that the code looks good. I haven't actually tested the functionality but I'm assuming that it all looks like it is working for you.

Besides the comments on the existing changes, please also update the Cranko metadata about version dependencies between the different packages in the repo. In practice this should be pretty straightforward:

- The `engine-vuex` module relies on new APIs introduced in the `engine` module, so the `internalDepVersions::@wwtelescope/engine` key in `engine-vuex/package.json` should be updated to a new value of the form `thiscommit:2022-04-14:TJBtju9`.
- The `research-app` module relies on new APIS introduced in both the `engine-vuex` and `research-app-messages` modules, so the respective `internalDepVersions` keys in `research-app/package.json` should be updated to analogously, with different random-ish strings after today's date.

These updates are required so that the next time we make a release, Cranko will be able to infer the correct version dependencies (i.e., the next release of `research-app` will require the next release of `engine-vuex`, and in fact we can't properly release `research-app` without releasing a new version of `engine-vuex` first.)

You can quickly generate requirement tags with the command `cranko show tctag` if you have Cranko installed, but you can also just generate them manually from today's date and some keyboard-mashing.

Carifio24:
I've made the requested changes and updated the metadata for Cranko.

pkgw (approved):

PR #196, opened 2022-07-16
Pin Vue to 2.6.x

Carifio24:
While building with the current `package.json`, one gets the following error with one of Vue's type declaration files:

```
> @wwtelescope/research-app-messages@0.0.0-dev.0 build /home/jon/dev/WWT/test/research-app-messages
> tsc

lerna ERR! npm run build exited 2 in '@wwtelescope/astro'
lerna ERR! npm run build stdout:

> @wwtelescope/astro@0.0.0-dev.0 build /home/jon/dev/WWT/test/astro
> tsc

../node_modules/vue/types/jsx.d.ts(39,7): error TS1110: Type expected.
../node_modules/vue/types/jsx.d.ts(39,20): error TS1005: ';' expected.
../node_modules/vue/types/jsx.d.ts(39,21): error TS1128: Declaration or statement expected.
../node_modules/vue/types/jsx.d.ts(40,1): error TS1128: Declaration or statement expected.

lerna ERR! npm run build stderr:
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! @wwtelescope/astro@0.0.0-dev.0 build: `tsc`
npm ERR! Exit status 2
npm ERR! 
npm ERR! Failed at the @wwtelescope/astro@0.0.0-dev.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/jon/.npm/_logs/2022-07-15T21_07_38_621Z-debug.log

lerna ERR! npm run build exited 2 in '@wwtelescope/astro'
lerna WARN complete Waiting for 2 child processes to exit. CTRL-C to exit immediately.
```

Running the lint script gives essentially the same error. This might be related to Vue 2.7, which was released 16 days ago. Pinning `vue` and `vue-template-compiler` to `~2.6.11` fixes this issue, so I propose we do that for now. This PR adds those pins.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,6 +25,7 @@
         "jsdom": "^16.5.0",
         "pako": "^1.0.11",
         "uuid": "^8.3.2",
+        "vue": "~2.6.11",
         "vue-color": "^2.8.1",
         "vue-select": "^3.11.2"
       },
@@ -60,7 +61,7 @@
         "uglify-js": "^3.9.1",
         "vue-class-component": "^7.2.3",
         "vue-property-decorator": "^8.4.1",
-        "vue-template-compiler": "^2.6.11",
+        "vue-template-compiler": "~2.6.11",
         "vuex-module-decorators": "^0.16.1"
       }
     },
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
         "@types/debounce": "^1.2.0",
         "@types/jsdom": "^16.2.13",
         "@types/vue-color": "^2.4.3",
-        "@types/vue-select": "^3.11.1",
+        "@types/vue-select": "3.16.0",
         "@wwtelescope/astro": "file:astro",
         "@wwtelescope/embed": "file:embed",
         "@wwtelescope/embed-common": "file:embed-common",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "@types/debounce": "^1.2.0",
     "@types/jsdom": "^16.2.13",
     "@types/vue-color": "^2.4.3",
-    "@types/vue-select": "^3.11.1",
+    "@types/vue-select": "3.16.0",
     "@wwtelescope/astro": "file:astro",
     "@wwtelescope/embed": "file:embed",
     "@wwtelescope/embed-common": "file:embed-common",
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "jsdom": "^16.5.0",
     "pako": "^1.0.11",
     "uuid": "^8.3.2",
+    "vue": "~2.6.11",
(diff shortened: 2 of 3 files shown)


pkgw:
Normally I'd say that I'd like to have a better understanding of where the error is coming from before pinning to an older version. But, looking at the release notes, it looks like Vue 2.7 is a backport of a lot of Vue 3 functionality to the 2.x series. This doesn't explain why things aren't working now, but I think it does mean that 2.8, 2.9, ... etc releases are unlikely. And if we're going to have to migrate, we should probably migrate to Vue 3 anyway. So all that is to say, in this particular case, I think I'm OK with pinning.

One random thought is that it might help to upgrade our TypeScript. I believe we have to pin to the TypeScript minor version and we're probably decently behind at this point. It would be nice to try not to fall too far behind the versions for these kinds of deps, although it can be time-consuming to stay on top of all of them.

Carifio24:
Yeah, my intent here was to pin it for the moment so that things are working until we figure out what the issue is. I think the TypeScript thought is a good one. It looks like TypeScript is at 4.7, while we're using 3.9 (that was the last 3.x release), so yeah, we're a bit behind. Do you remember the reason for the TS pin?

pkgw:
No, the content updates shouldn't have affected any of this stuff.

Now there seems to be a different failure:

```
> @wwtelescope/astro@0.2.1+20220718 build
> tsc

../node_modules/@types/vue-select/index.d.ts(241,46): error TS2314: Generic type 'ExtendedVue' requires 5 type argument(s).
```

Chaos!

Carifio24:
Yeah, I would've been surprised if they did, but worth a check. It seems like the BrowserStack test was failing because the parser wasn't set to parse XML. I don't know why this worked without this setting in the past, but I've updated the `JSDOM` options to specify XML.

Now we're getting errors because of a just-released 3.16.1 update (as in, 3 hours ago) to `@types/vue-select`. It looks like one of the changes to this version is in response to a change in the `ExtendedVue` type in Vue 2.7 (clearly we're not the only one dealing with these sorts of issues). Pinning `@types/vue-select` to 3.16.0 makes everything work for me locally, so that's one possible option. I'm not a huge fan of adding another restrictive pin, but I think otherwise we'll need to update TypeScript, which feels like a bigger change.

pkgw:
Yeah, let's just pin that too. It will be worth adding some extra description in the Git commit message so that we can remind ourselves why the pin is there, but I'm not optimistic that this stuff will all be ironed out especially soon, and we need to get our builds going.

Carifio24:
Ok, I've added the pin, and all the checks are good now!

Issue #199, opened 2022-08-01
Firefox/Linux problems with "WebGL context was lost"


DougBurke:
I have occasional issues with this. At best I can come up with is that this happens the first time I'm accessing the WWT, as I can reload the page and it goes away. My current firefox is 102.0.1 installed on ubuntu via a snap.

This is a new version of the CSC stress test (as it's in prep for the 2.1 release, so more polygons and even-more circles):

https://cxc.cfa.harvard.edu/csc2/wwt21.html

For the user the symptoms is that the page appears to load but either the stack polygons are not shown, or nothing from the WWT is shown. If I use the console log then at some point I see the message

```
WebGL context was lost.
```

and then at a later point it shows an error, such as

```
Uncaught TypeError: WebGL2RenderingContext.shaderSource: Argument 1 is not an object.
```

Unfortunately I don't think I (as the author) can catch or identify this problem, as it's coming from the WWT code (unfortunately I don't have an example of the problem to show file name/location info), so I don't know how I can let the user know they should try a reload (or if there's some other issue).

[I'm going to be away for a while, so I just wanted to note this in case it's a known problem or has an obvious solution]

pkgw:
Thanks for reporting! I've actually been looking into this issue recently for a different reason: I find that it often affects me on Firefox/Android, resulting in a blank screen in our apps. Not great!

I don't understand the issues too well, but I *think* that basically we need the web engine to notice "context lost" events and respond to them by re-acquiring it. As I understand it, if you do it right, there might be a hiccup in rendering but basically the app will be able to keep on working without requiring reloads or anything. At least, sometimes — I think this can also happen when
you run out of memory on the GPU, and stuff like that, in which case I think you also have to reduce the complexity of the scene that you're trying to render if you don't want to get stuck in an unending loop of context-lost events.

When the context is lost, most WebGL routines become no-ops, but the `createShader()` routine starts returning null, which leads to the TypeError you pasted above. So we'll also need to add some try/catches that basically swallow these errors so that the JS can keep on running.

CC @imbasimba, I know you're busy, but is this an aspect of WebGL that you've dealt with before?

DougBurke:
Well,I'm glad it's not just a problem with me ;-)

imbasimba:
[...]

PR #200, opened 2022-08-10
WebGL context loss / restored

imbasimba:
The prototype implementation gracefully handles loss of WebGL context, as well as properly restores all WebGL variables for Toast Tiles. Something along these lines is what is needed for the rest of the application to properly recover from a context loss.

- [ ] Loss of context can happen at any time, so the shader's `if (gl.isContextLost())` check is not enough. Use try/catch instead.
- [ ] Fix all committed TODO's.
- [ ] Handle context loss & restored for the rest of the application.

Fully implementing this would let WWT gracefully recover from loss of WebGL context, and not require a page reload like the one discussed here: https://github.com/WorldWideTelescope/wwt-webgl-engine/issues/199.

This PR does not address any potential issues which may cause a loss of context in the first place.

### DIFF
--- a/engine-vuex/src/Component.vue
+++ b/engine-vuex/src/Component.vue
@@ -65,6 +65,27 @@ export default class WWTComponent extends Vue {
       this.internalUpdate();
     };
 
+    const wwtcmpt0 = document.getElementById("wwtcmpt0");
+    if (wwtcmpt0 != null){
+      const wwtCanvas = wwtcmpt0.children[0];
+      if (wwtCanvas != null) {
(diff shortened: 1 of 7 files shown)


pkgw:
I'm afraid that this code will need to be ported over to the pure-JS implementation (although hopefully that won't be so hard, the translation is pretty literal). Seeing as this PR has been idle for a while, I'm going to close it now. For what it's worth, I've *felt* like I've noticed this issue a lot less lately ... it definitely feels like it isn't happening as much on my phone (Firefox/Android), which is where I was noticing it before.

PR #204, opened 2022-08-24
Fix bug in time span parsing

### LINKED ISSUE #54: Slides with 1+min duration are recognized as 10+ min
When making a tour, slides that are 59 sec operate at the desired duration. At 1 min, slides are recognized as a duration in the web client as 10 min instead of 1 and this scales consistently: if the slide is 2 min (recognized as 20) or 4 min (recognized as 40). This is a web client specific issue.

Carifio24:
This PR resolves #54. The tour duration bug described in that issue is due to extra factors of 10 in two constants in the engine's `Util::ParseTimeSpan` static method. This function converts an HH:MM:SS.S time duration to milliseconds. However, the conversion factors for minutes and hours are both 10 times larger than they should be - e.g. there are 60000 milliseconds in one minute, not 600000. The same issue occurs with hours.

### DIFF
--- a/engine/wwtlib/Util.cs
+++ b/engine/wwtlib/Util.cs
@@ -105,8 +105,8 @@ namespace wwtlib
 
             if (parts.Length == 3)
             {
-                val += int.Parse(parts[0]) * 36000000;
-                val += int.Parse(parts[1]) * 600000;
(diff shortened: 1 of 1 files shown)


pkgw:
Great catch!

PR #211, opened 2022-09-06
Corrected half pixel off issue when displaying standalone FITS files

imbasimba:
This fixes the issue of half pixel off for untiled FITS. 

Long story short, the issue is that `CRPIX` is referring to the center of the pixel, while `referenceX` is not. 

Does not fix issue with half pixel off for tiled tangential FITS - I'll look into that issue tomorrow.


pkgw:
This aligns with the code that we have in `wwt_data_formats.imageset.Imageset.set_position_from_wcs()`, so yes, I think this is correct.

Browserstack failure appears to be spurious.

PR #212, opened 2022-09-08
FITS position adjustments

imbasimba:
This PR has two parts:
1. Zoom position was wrong for both tiled tangential imagesets and untiled images
2. Half pixel correction for tiled tangential imagesets

**Regarding the first issue:**
For untiled images, `GotoRADecZoom` could be called with a RA < 0, which got clamped to 0, resulting in a wrong position. For tiled tangential imagesets the goto destination was off by the `offsetX` and `offsetY` in the wrong directions. Tbh, I did not look into the exact reason because we need to decide the desired behaviour of the goto action. Right now, we have a mix of solutions, where we sometimes try to go to the center of the image and sometimes to the center of the coordinate system like `CRPIX`. The buttons in the research app seem to always go for the coordinate system position, which arguably is more correct. For now I just removed the `offsetXY`, to get the same goto position inside the engine as in the research app.  Do you think this is the behavior we should converge towards?

The current status with this PR is that all `TilingMethods` (UNTILED, TOAST, TAN, HIPS) goto `CRPIX` if the goto button is pressed inside the research app. The same position is the destination for the automatic engine goto onload for TOAST and TAN. For UNTILED and HIPS, the positions are still the center of the image, so whatever way we go, we should probably align the rest.

**Regarding the second issue:**
This is a tricky one... It is very easy to take a mental misstep here, so I'd highly appreciate a second pair of eyes. Did I get it completely right by flipping the sign in the y axis for the bottoms up etc? What I have here seems to properly handle all the test cases I've thrown at it.

Just like https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/211 we want the pixel center to be the defining feature of the position and not the corner of the pixel. Since the tiled tangential imagesets has levels, we have to make sure to adjust all levels by a half pixel of the _original image_, which in this case is the pixel size of the bottom level.

Here is some handy pywwt code to test different variations:
```
from pywwt.jupyter import connect_to_app
wwt = await connect_to_app().becomes_ready()
import numpy as np
from toasty import TilingMethod
from astropy.wcs import WCS

size = 256 * 16 + 1 # size <  257 results in a SkyImage, which is not yet displayed correctly
array = np.arange(size * size, dtype=float).reshape(size, size)

array[size - 1][size - 1] = 0
array[0][0] = size * size
array[0][size - 1] = size * size
array[size - 1][0] = 0
array[int((size - 1) / 2)][int((size - 1) / 2)] = size * size

wcs = WCS()
wcs.wcs.ctype = 'RA---TAN', 'DEC--TAN'
wcs.wcs.crval = 0, 0
wcs.wcs.crpix = size, size
wcs.wcs.cdelt = 20 / size, 20 / size
wcs._naxis = [size, size]

wwt.layers.add_image_layer(image=(array, wcs), tiling_method=TilingMethod.TAN, override=True, name='tan')
#wwt.layers.add_image_layer(image=(array, wcs), tiling_method=TilingMethod.UNTILED, override=True, name='untiled')
#wwt.layers.add_image_layer(image=(array, wcs), tiling_method=TilingMethod.HIPS, override=True, name='HiPS')
#wwt.layers.add_image_layer(image=(array, wcs), tiling_method=TilingMethod.TOAST, override=True, name='TOAST')
```
There are still some discrepancies at the other corners, which could be investigated later.

### DIFF
--- a/engine/wwtlib/Imageset.cs
+++ b/engine/wwtlib/Imageset.cs
@@ -841,7 +841,7 @@ namespace wwtlib
                 }
                 if (Levels > 0)
                 {
-                    return CenterX - OffsetX;
+                    return CenterX;
                 }
                 else
                 {
@@ -860,7 +860,7 @@ namespace wwtlib
                 }
                 if (Levels > 0)
                 {
-                    return CenterY + OffsetY;
+                    return CenterY;
                 }
                 else
                 {
--- a/engine/wwtlib/Imageset.cs
+++ b/engine/wwtlib/Imageset.cs
@@ -832,42 +832,83 @@ namespace wwtlib
             }
         }
 
+        // Calculate either the X or Y coordinate of the estimated image center.
+        //
+        // This estimate has some important limitations. First, because images
+        // might contain transparent regions, the "center" of the image that a
+        // user will perceive might have nothing to do with the center of the
+        // image bitmap. For instance, imagine that the bitmap is 100x100 but
+        // that everything is transparent except for 10x10 pixels in the
+        // top-left corner. We don't know anything about the "barycenter" of the
+        // image here, so we can't account for that.
+        //
+        // Second, for untiled SkyImage imagesets, to properly compute the
+        // bitmap center we need its dimensions, which simply aren't available
+        // here. All we can do is guess a "reasonable" image size.
+        //
+        // For these reasons, this method should be avoided when possible. The
+        // preferred way to "know" the location of an image's center is to wrap
+        // the image in a Place object, which can just specify the exact
+        // coordinates and zoom level too.
+        //
+        // Even disregarding the above, it's non-trivial to locate the image
+        // center because of the OffsetX/Y parameters and potential rotation of
+        // the image's coordinate system relative to the sky.
+        private double CalcViewCenterCoordinate(bool isX)
+        {
+            double rot = Coordinates.DegreesToRadians(rotation);
+            double crot = Math.Cos(rot);
+            double srot = Math.Sin(rot);
+
+            double dx = 0, dy = 0;
+
+            if (Levels > 0) {
+                dx = -offsetX;
+                dy = offsetY;
+            } else {
+                // This is the part where we need the image's dimensions to
+                // be able to compute the center coordinate correctly. Since
+                // we don't have that information, we just guess :-(
+                double effWidth = 800;
+                double effHeight = 800;
+
+                dx = (offsetX - effWidth / 2) * baseTileDegrees;
+                dy = (effHeight / 2 - offsetY) * baseTileDegrees;
+            }
+
+            if (bottomsUp) {
+                dx = -dx;
+            }
+
+            if (isX) {
+                return centerX + dx * crot + dy * srot;
+            } else {
+                return centerY - dx * srot + dy * crot;
+            }
+        }
+
         public double ViewCenterX
         {
             get {
-                if (WcsImage != null)
-                {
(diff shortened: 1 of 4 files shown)


engine/wwtlib/Imageset.cs
```diff
@@ -841,7 +841,7 @@ public double ViewCenterX
                 }
                 if (Levels > 0)
                 {
-                    return CenterX - OffsetX;
+                    return CenterX;
```
pkgw, line 844: I think that we should continue including OffsetX here. The best thing that we can do is assume (hope) that the visual center of the image is around the center of the center of the top-level tile, which is found at CenterX offset by OffsetX.

If I understood the PR comment correctly, though, you said that the sign being applied here was wrong. Based on [the documentation I wrote][1] I *think* that it should be correct as-is ... but I wrote both things so that's not very meaningful.

[1]: https://wwt-data-formats.readthedocs.io/en/latest/api/wwt_data_formats.imageset.ImageSet.html#wwt_data_formats.imageset.ImageSet.offset_x

engine/wwtlib/Imageset.cs
```diff
@@ -860,7 +860,7 @@ public double ViewCenterY
                 }
                 if (Levels > 0)
                 {
-                    return CenterY + OffsetY;
+                    return CenterY;
```
pkgw, line 863: Same commentary as in `OffsetX`. Here too, based on my own documentation, I think that the sign should be correct.

pkgw (commented):
My comment on the centering issue below.

From your description, I think that the research app is the thing that needs changing, not the engine, modulo the sign error that may be present.

engine/wwtlib/Imageset.cs
imbasimba: Hmm... Yeah for the tiled tangential test case posted above the signs definitely needs to be flipped to put the camera at the center of the image. `+ OffsetX` and `- OffsetY` is correct in this case. The documentation makes me question whether flipping the sign would mess up other cases though... Are we maybe doing something speical since it is FITS files with bottomsUp etc?

pkgw:
I am finally taking a look at this in detail, starting with the ViewCenter computation. I realized that we had a big oversight there: the rotation of the image coordinate system needs to be taken into account! I think that explains why you were seeing different sign conventions for handling the offset parameters. I've worked through a series of examples and am now pretty confident that I have the right computation.

engine/wwtlib/Imageset.cs
pkgw: OK, I think our disagreement must have been due to a need to consider the image rotation too, as in my follow-up comment.

pkgw:
Now digging into the half-pixel correction. Your test case does a very nice job of demonstrating that there are issues! But I don't believe that TangentTile is the place to apply the fix; as it is, I believe that your code will induce little one-pixel overlaps along all of the tile edges of the interior of large image, which can't be right. I think the issue is at the stage in which we convert from WCS to WWT's internal coordinate representation.

pkgw:
OK, I've been continuing to work on this.

The really glaring issues that I noticed before seem to be due to the fact that the test case spans such a large angular area. The tangential approximations break down badly in that case.

If I change the `cdelt` setting to 1e-5, that basically removes those issues from what we're trying to looking at here. In that case, I do see a half-pixel offset with the TAN projection ... *when the image dimension is odd*. There is an issue where the `wwt_data_formats` WCS-to-WWT conversion, which is not fully in sync with the way that `toasty` is tiling things. I have a fix for `wwt_data_formats` that makes the problem go away.

PR #217, opened 2022-11-17
Move from Lerna to Yarn workspaces

Carifio24:
This PR moves our monorepo management from Lerna to [Yarn workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/). I've only made the most obvious CI changes, so there will inevitably be issues to work out, but the current form of this should be the basis for what we want. @pkgw There are some Cranko steps in the CI release workflows that I'm sure will require your expertise.

The conversion here is relatively straightforward. We ditch `lerna.json` and `package-lock.json` in favor of the Yarn lockfile `yarn.lock` and the configuration `.yarnrc.yml`. (Yarn suggests deleting `package-lock.json` if using yarn; see the disclaimer at the end of [this article](https://classic.yarnpkg.com/blog/2018/06/04/yarn-import-package-lock/)). We specify the workspaces we want in the top-level `package.json` file. The sub-packages get changed from `file` to `link` protocols there so that Yarn will symlink to the source directories inside of `node_modules`.

Yarn does have a [Zero-Installs](https://yarnpkg.com/features/zero-installs) feature, which they define as: "In order to make a project zero-install, you must be able to use it as soon as you clone it." This involves checking in the Yarn cache and a `.pnp.cjs` file that contains the dependency tree for Node. This PR doesn't take this approach - it feels like a pretty significant change to me. Potentially something to look into, though.


Carifio24:
Rebased to remove the package-lock.json merge conflict

pkgw:
OK, current error looks to be related to something I was going to ask about anyway:

```
Error: Cannot find module '/home/vsts/work/1/s/.yarn/releases/yarn-3.3.0.cjs'
```

It looks like the metadata files are basically pinning the build to use Yarn 3.3.0. Is that the recommended approach these days? I believe that the CI images will come with Yarn preinstalled, and its version will probably update pretty frequently, so my initial feeling is that we should just be flexible about using whatever `yarn` command is available. But maybe the more modern style is to be stricter about things?

Carifio24:
It looks like this got added to the yarnrc when I explicitly set the yarn version inside my local folder. I don't think we want this - if we we doing the zero-installs I could see keeping it, but we aren't committing `.yarn`.

pkgw:
Progress! Maybe the packages in the `workspaces` list need to be listed in a topological sort order?

Carifio24:
Okay, yeah, I can reproduce with a clean checkout as well - previous tinkering had made that not be the case in my local version before. I'll take a look at the JupyterLab workspace and see what ideas I can glean from there for getting the build order correct.

Carifio24:
So the key was to add the `-t` option to the `yarn workspaces foreach` commands - "t" standing for topological.

Carifio24:
So the commit history has gotten a bit ugly here (time for some history rewriting?), but the checks are all passing. I'm not yet sure what, if anything, needs to be updated for any release/publish steps.

Carifio24:
Ok, the commit history has been cleaned up a bit

pkgw:
OK, I think this is all together now! I've made a pass to homogenize the `tsconfig.json` files and move all of the dependencies from the toplevel `package.json` file into their individual packages. This is a little bit more verbose, but it's more correct for each package to declare its actual dependencies — and Yarn is smart enough to emit things in the toplevel `node_modules` directory whenever possible, so it's not like we get 10 copies of TypeScript installed. This means that now commands like `cd engine && yarn build` work now, which is great. I was also able to fix things like missing peer dependencies.

I also took a pass through the CI files and I think that things should be in good shape with them now.

Finally, I updated the README to reflect all of these changes. I think it's the only piece of documentation that goes into the details of the build process.

pkgw:
I attempted a first release, and it looks like the publish-to-NPM step will need some attention — but otherwise, things are looking good. I'll probably push some commits directly to `master` as I work on the publishing step since using PRs doesn't actually help test the functionality, and just adds a bunch of latency to the process.

pkgw:
All right, it looks like the release automation is working again!

PR #218, opened 2022-11-18
Ensure that source message doesn't try to send proxy

Carifio24:
This PR fixes a small bug from the Vue 3 migration. The research app component's `lastSelectedCount` is now turned into a `Proxy`, which causes a `DOMException` when sending the message describing the last selected source (things work fine app-side). The fix here is to use Vue's `isProxy` and `toRaw`, if necessary, to serialize the original non-proxy object.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,7 +55,7 @@
         "nightwatch": "^2.3.9",
         "sass": "^1.26.3",
         "sass-loader": "^13.0.2",
-        "typedoc": "^0.23.19",
+        "typedoc": "0.23.20",
         "typescript": "^4.8.4",
         "uglify-js": "^3.9.1",
         "vue-template-compiler": "~2.6.11"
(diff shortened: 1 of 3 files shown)


pkgw:
Hmm, doc build error appears to be a real error, although I'd be shocked if it had anything to do with the change introduced here:

```
> @wwtelescope/engine-pinia@0.0.0+20221118 doc
> typedoc src/index.ts

lerna ERR! npm run doc stderr:
../node_modules/pinia/dist/pinia.d.ts:576:132 - error TS2339: Property 'suffix' does not exist on type 'MapStoresCustomization'.

576 export declare function setMapStoreSuffix(suffix: MapStoresCustomization extends Record<'suffix', string> ? MapStoresCustomization['suffix'] : string): void;
```

I don't suppose you encountered anything like this while working on the port?

Carifio24:
This is a new error to me. It seems to be a bad interaction between TypeDoc, Pinia, and TypeScript, all of which have a new release in the past 5 days. I'll take a look into this.

Carifio24:
So the Pinia file that defines `MapStoresCustomization` hasn't been changed since May, so I don't think that's the offender here (though it [doesn't actually define](https://github.com/vuejs/pinia/blob/v2/packages/pinia/src/mapHelpers.ts#L17) the `suffix` member the docstring says it does). Bumping TypeDoc up to 0.23.21 moved TypeScript from 4.8 -> 4.9, so maybe the issue lies there? (Keeping TypeDoc at 0.23.20 resolves the issue)

pkgw:
Probably something like that. Now that we're tracking more recent deps we'll probably run into more of this kind of thing — the joys of the fast-moving ecosystem. I think it's fine to keep TypeDoc and TypeScript dialed back for now; the thing I always worry about is how/where to document the reason for doing so, and hopefully providing outselves a little reminder to revisit in a month or whatever.

Carifio24:
Pinning `typedoc` and `typescript` versions did the trick. As for a good way to document this and prod revisiting, I'm not sure. I guess putting this info in the git commit would be useful if someone already knows where to look, but won't help on the reminding front. Obviously we can't add comments in `package.json`, but we could always do something like [this](https://stackoverflow.com/a/58385464) to add "comment" fields to the JSON object.

PR #220, opened 2022-11-28
Allow changing selection proximity in research app

Carifio24:
This PR allows changing how close the pointer needs to be (in pixels) for a source to be considered close enough to select (by default, this is 4 pixels). This is done by adding a `selectionProximity` setting to the research app, which allows this to be changed via a `modify_settings` message. The `distanceThreshold` data field in the research app component has also been renamed to `selectionProximity` for consistency (overall, I felt that "selection proximity" gave a more obvious meaning, but I'm open to other suggestions).

### DIFF
--- a/research-app-messages/src/settings.ts
+++ b/research-app-messages/src/settings.ts
@@ -38,14 +38,16 @@ export function isGenericSetting(obj: any): obj is GenericSetting {  // eslint-d
 /** Settings for the overall application.
  *
  * This should be thought of as an enumeration type, all of whose options are
- * compatible with [[GenericSetting]]. Right now, there is only one option
+ * compatible with [[GenericSetting]]. Right now, there are only two options
  * available, though.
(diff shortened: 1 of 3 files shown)


pkgw:
Looks great. My only comment is that since research-app is relying on a new interface added to the research-app-messages module, its `internalDepVersions` field in `package.json` needs to be updated with a Cranko `thiscommit` thing. This will probably require rebasing the PR onto the latest `master` since I've just merged in the big Yarn rework, which touched all of the `package.json` files.

Carifio24:
Just rebased onto the new master + added the tctag for `research-app-messages`.

Carifio24:
BrowserStack is failing, with the error being a failed git fast-forward during the "Restore release commit" step. Could this be due to the fact that there have been new releases of some of the packages since the build steps here originally ran?

PR #222, opened 2023-01-16
Expose getting time for some goto movements

Carifio24:
This PR adds functionality to the engine to allow obtaining the time necessary for the `GotoRADecZoom` and `GotoTargetFull` motions. This is done via the functions `TimeToRADecZoom` and `TimeToTargetFull`, respectively, along with some refactoring of shared calculations. `TimeToRADecZoom` is also exposed to the Pinia layer - I didn't do this for `TimeToTargetFull` since `GotoTargetFull` isn't exposed.

These implementations are based on the implementations of the corresponding goto functions. A few notes/thoughts:
• I did some testing in the browser, by comparing the reported and actual execution times (nothing fancy - I just surrounded an awaited goto call by `Date.now`s), and agreement is very good - on average, I found the execution time to be about 0.1% more than the reported time.
• Since some of the intermediate values in `ViewMoverSlew` are retained for later use, I determined that the best thing to do was to just create an instance and use its move time.
• `ViewMoverSlew` reports its time in seconds, so I've retained that. But maybe milliseconds would be more natural?
• It would be nice to do the same thing for `GotoTarget`. That's a much more complicated function, so I'll save that for a future PR.

### DIFF
--- a/engine-helpers/src/index.ts
+++ b/engine-helpers/src/index.ts
@@ -467,6 +467,27 @@ export class WWTInstance {
     return this.makeArrivePromise(instant);
   }
 
+  /** Returns how long moving to the given position will take, in seconds.
+   *
+   * This wraps the underlying engine function of the same name, but homogenizing some
+   * of the angular arguments to use radians.
+   *
+   * @param raRad The RA to seek to, in radians
+   * @param decRad The declination to seek to, in radians
+   * @param zoomDeg The zoom setting, in *degrees*
+   * @param rollRad If specified, the roll of the target camera position, in radians
+   * @returns A void promise that resolves when the camera arrives at the target position.
+   */
+   timeToRADecZoom(raRad: number, decRad: number, zoomDeg: number, rollRad?: number): number {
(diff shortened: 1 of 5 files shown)


engine-helpers/src/index.ts
```diff
@@ -467,6 +467,27 @@ export class WWTInstance {
...
+   * @returns The amount of time, in seconds, that moving to the given position would take.
+   */
+   timeToRADecZoom(raRad: number, decRad: number, zoomDeg: number, rollRad?: number): number {
+    const time = this.ctl.timeToRADecZoom(
+      raRad * R2H,
+      decRad * R2D,
+      zoomDeg,
+      rollRad
+    );
+    return time;
```
pkgw, line 488: Not worth changing, but this could just be `return this.ctl.timeToRADecZoom(...)`

pkgw (approved):
Looks great! Would be nice to have some associated tests but we just don't have the infrastructure for that right now, so that's out of scope for this PR.

PR #223, opened 2023-01-19
Add store functionality for accessing general and imageset layers

Carifio24:
This PR adds functionality to the Pinia store to allow straightforward access to general and imageset layers, in the same way that we currently have access to spreadsheet layers.

Additionally, this also does some refactoring of these layer access functions. We currently have utility functions defined external to the store for spreadsheet layer access. The motivation behind this setup was to allow reuse between Vuex actions and mutations. Since mutations aren't a thing in Pinia, we no longer need this separation. Additionally, the general layer access function added in this PR is used to avoid repeating code inside the more specific access functions.

### DIFF
--- a/engine-pinia/src/store.ts
+++ b/engine-pinia/src/store.ts
@@ -24,6 +24,7 @@ import {
   Imageset,
   ImageSetLayer,
   InViewReturnMessage,
+  Layer,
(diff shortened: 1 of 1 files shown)


pkgw (approved):
Looks great!

PR #228, opened 2023-02-03
Use correct image for point and circle modes

Carifio24:
This PR modifies spreadsheet and VO table layers to use the correct images when the plot mode is `point` or `square`. Previously these modes were using an incorrect marker.

The reason for this is that, with the exception of the circle, the images for the layer markers come from [here](https://web.wwtassets.org/engine/assets/pins.png). Each sub-image here is 32 pixels square. The appropriate image is accessed via an index, which is zero-indexed and runs through the grid in reading order. This index is converted into a row and column, and then a 32 pixel square starting from the appropriate point is used for the image texture. However, the circle and square cases were using the wrong index, so the wrong image was being used. This PR fixes those indices.

I suppose this is a breaking change, in the sense that anyone who relied on this incorrect behavior will now see something different be displayed. But I think there's a benefit here of both allowing access to more plot types, and just having options be correct.

### DIFF
--- a/engine/wwtlib/Layers/SpreadSheetLayer.cs
+++ b/engine/wwtlib/Layers/SpreadSheetLayer.cs
@@ -2678,10 +2678,10 @@ namespace wwtlib
                         pointList.DrawTextured(renderContext, CircleTexture.Texture2d, opacity * Opacity);
                         break;
                     case PlotTypes.Point:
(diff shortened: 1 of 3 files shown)


Carifio24:
I've also exposed these two plot types as options in the research app UI for spreadsheet layers (which includes HiPS catalogs).

PR #231, opened 2023-02-19
Expose roll angle in `ViewStateMessage`

Carifio24:
This PR modifies the research app and `ViewStateMessage` to expose the roll angle in the application state updates. The intended downstream use case is for a client like `glue-wwt` or CosmicDS, which may want to save and restore the state of the WWT view. For consistency with [`CenterOnCoordinatesMessage`](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/research-app-messages/src/classic_pywwt.ts#L115), I decided to expose this number in degrees.


pkgw:
Looks good!

PR #232, opened 2023-02-23
Allow importing engine in non-browser context

Carifio24:
This PR adds the minimal set of guards necessary to import the engine package in a non-browser context. The intended use case for this is for server-side code that doesn't need the browser APIs. We need these guards because there are a few objects that get instantiated on the module load (e.g. `URLHelpers.singleton`) that call browser APIs.

FWIW, I originally tried using some ScriptSharp functionality (see [this branch](https://github.com/Carifio24/wwt-webgl-engine/tree/ssr-guards-v1)), but that didn't help - it looks like the `typeof` check needs to be in the same scope as the browser API use.

### DIFF
--- a/engine/wwtlib/Graphics/Texture.cs
+++ b/engine/wwtlib/Graphics/Texture.cs
@@ -61,6 +61,7 @@ namespace wwtlib
         public void Load(string url)
         {
             URL = url;
+            Script.Literal("if (typeof document === \"undefined\") { return; }");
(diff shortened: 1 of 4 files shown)


pkgw:
I presume that you've tested this in the Constellations SSR context and it seems to solve the problems you were having?

Carifio24:
Yeah, that's been my test case for these changes.

PR #234, opened 2023-03-03
Update BrowserStack configurations

Carifio24:
While setting up a BrowserStack configuration for the MiniDS repo, I noticed that BrowserStack has seemingly updated the format of the field names in their configurations. We had been using snake_case, but they seem to now want camelCase: see [here](https://www.browserstack.com/docs/automate/selenium/select-browsers-and-devices) (note that Nightwatch uses Selenium). This PR makes those updates. Looking back at our most recent CI logs, it seems that only the default browser environment was running, rather than the several specified in the workflow, which I suspect is related to this. I also fixed a few minor things that my linter was complaining about.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ To run the test suite on a local version of the app with ChromeDriver:
 yarn run serve-research &  # or equivalent
 cd tests
 yarn build
-yarn run local
+yarn local
 ```
 
 To run tests on a local version of the app via BrowserStack's multi-browser,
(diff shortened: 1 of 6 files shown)


Carifio24:
So it was actually the `--` in the run command (needed for npm, but not yarn) that was preventing the other environments from running. I suspect that was preventing us from seeing issues from this, though.

pkgw:
Thanks for straightening this out!

PR #235, opened 2023-03-10
Generalize and expose capture functionality

Carifio24:
This PR generalizes the current `CaptureThumbnail` functionality with a new `CaptureFrame` method of the `WWTControl` class. This method captures the current frame of the WWT view, and allows specifying the width, height, and format of the image output. `CaptureThumbnail` now simply calls the more general function. Additionally, this PR exposes this method to the Vue layer via the Pinia store.

Internally, the frame capture uses `toBlob`, which uses a `BlobCallback` to allow manipulation of the created image blob. Since our API is Promise-based, the exposure to the store is done via a Promise that resolves to the blob in the callback.

### DIFF
--- a/engine-helpers/src/index.ts
+++ b/engine-helpers/src/index.ts
@@ -303,6 +303,18 @@ export interface SetupForImagesetOptions {
   background?: Imageset;
 }
 
+/** Options for [CaptureFrame] */
+export interface CaptureFrameOptions {
+  /** The desired image width. */
+  width: number;
+
+  /** The desired image height. */
+  height: number;
+
+  /** The desired image format (e.g. `"image/jpeg"`). */
+  format: string;
+}
+
 interface ResolveFunction<T> {
   (value: T): void;
 }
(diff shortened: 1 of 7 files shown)


engine-helpers/src/index.ts
```diff
@@ -303,6 +303,18 @@ export interface SetupForImagesetOptions {
...
 
+/** Options for [CaptureFrame] */
+export interface CaptureFrameOptions {
+  /** The desired image width. */
+  width: number;
+
+  /** The desired image height. */
+  height: number;
+
+  /** The desired image format (e.g. `"image/jpeg"`). */
```
pkgw, line 314: We should explicitly specify that the value of this parameter should be a MIME type, if that's indeed the case. With string-valued parameters it's always good to specify the desired format of the string in as much detail as possible.

(By the same token, we should specify that the width and height are in pixels, here.)

pkgw (commented):
Looks good! I only see a few things to tweak. Besides the comment below, it looks like there need to be some updates to Cranko internal dependency requirements because there are new APIs that are being used across packages here.

engine-helpers/src/index.ts
Carifio24: Good call! I've updated the descriptions for these fields.

pkgw: Did you see my other comment about the Cranko updates?

Carifio24: I somehow missed that before! Tags have been updated

PR #239, opened 2023-03-23
Add video capture as frames functionality

Carifio24:
This PR is my proposal for adding the "video capture as frames" functionality, which exists in the Windows client, into the web engine. The C# code for handling this is largely stolen from the Windows client implementation, with a few small tweaks to fit this context.

For the sake of completeness, the basic idea is that this adds a notion of being in a "frame dumping" state to the engine. The main change is that this affects how the `SpaceTimeController` updates its current time, which is now controlled by a new `metaNow` member value. If we aren't frame dumping, things update as normal. But if we are, setting this value via the setter is a no-op - to move time forward we need to call `NextFrame`, which increments time by the appropriate amount for the frame dumping FPS. This method is called when the tile cache is empty (or if we set the `WaitDownload = false` in the frame dumping parameters). If that's not the case, we continue to process the queue and re-check during the next `RenderOneFrame` call.

The other question that this addresses is how what to do with the created `Blob` objects that represent the captured frames. This question has an obvious answer on the desktop - save to the specified directory with an appropriate filename. That's obviously not an option here, so the approach this takes is to expose a [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams) of frame blobs that the client can handle however it likes. The implementation is relatively straightforward - the `BlobCallback` passed into the engine adds the blob into the stream. When the number of received blobs meets the specified total number of frames, the stream is closed (engine-side, this `BlobReady` is set to `null`). It seems a bit redundant to keep track of frames both engine- and TS-side, but I think we need to. I initially tried passing in an "on capture complete" callback to close the stream, but the issue is that when the engine captures the last frame and when the stream receives it will be different, and this will lead to the stream being closed prematurely.

### DIFF
--- a/engine-helpers/src/index.ts
+++ b/engine-helpers/src/index.ts
@@ -1048,13 +1048,13 @@ export class WWTInstance {
     const videoStream = new ReadableStream<Blob | null>({
       start(controller: ReadableStreamDefaultController) {
         function stream() {
+          let received = 0;
           wwtControl.captureVideo(blob => {
-              console.log(blob);
+              received++;
               controller.enqueue(blob);
-            },
-            () => {
-              controller.close();
-              console.log("Stream closed!");
+              if (received >= options.totalFrames) {
+                controller.close();
+              }
             },
             options.name,
             options.width,
--- a/engine-helpers/src/index.ts
+++ b/engine-helpers/src/index.ts
@@ -315,6 +315,26 @@ export interface CaptureFrameOptions {
   format: string;
 }
 
+export interface CaptureVideoOptions {
+  /** The video name */
+  name: string;
+
+  /** The desired image width, in pixels. */
+  width: number;
+
+  /** The desired image height, in pixels. */
+  height: number;
+
+  /** The number of frames per second */
+  framesPerSecond: number;
+
+  /** The total number of frames to capture */
+  totalFrames: number;
+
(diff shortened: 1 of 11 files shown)


engine-helpers/src/index.ts
```diff
@@ -1021,4 +1041,34 @@ export class WWTInstance {
         options.format);
     });
   }
+
+  /** Capture a video as a sequence of frames using the given parameters
+   *
+   * This function returns a readable stream whose values are the exported frames.
+  */
+  captureVideo(options: CaptureVideoOptions): ReadableStream<Blob | null> {
```
pkgw, line 1049: Does the type yielded here need to be `Blob | null` or can it just be `Blob`? Does the stream yield `null` once it's closed, or something?

pkgw (commented):
Looks very nice! I just have one question below.

engine-helpers/src/index.ts
Carifio24: It's `Blob | null` to match `BlobCallback` - the argument to the callback in `HTMLCanvasElement::toBlob`, which is where the blob-relevant code ultimately gets executed, is `null` if the image can't be created for some reason.

Carifio24:
I pushed one more commit updating the commit tags for Cranko.

PR #258, opened 2023-06-27
Change grid color settings to be strings

Carifio24:
In #226 the ability to change the colors of the various grid overlays was added to the engine. Inside the engine this works great, but I noticed an issue when attempting to expose these color options to `pywwt`.

The issue is that these new grid color settings expect a WWT `Color` object, but the pywwt's Color traitlet sends over a string, meaning that it can't be used to change the grid color. After looking into this a bit more, I realized that the way that I added in the color settings in #226 was inconsistent with the existing color settings (such as the colors for drawing constellations, see [here](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Settings.cs#L95)).

The approach this PR takes is to change the grid colors to be represented by strings, and then use `Color.Load` to get an actual `Color` instance for drawing the grids (similar to what is done e.g. [here](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Constellations.cs#L288)). This required relatively minimal code changes, and is consistent with how the constellation setting colors are defined.

The alternate approach would be to leave these settings as colors, and add some logic somewhere to parse the string into a color when it gets assigned to the relevant setting. Since C# doesn't let property setters accept values that differ from the property's type, this felt more involved to me.

### DIFF
--- a/engine-types/src/index.ts
+++ b/engine-types/src/index.ts
@@ -261,6 +261,7 @@ export interface Thumbnail {
  * the engine itself. */
 export type BaseEngineSetting =
   ["actualPlanetScale", boolean] |
+  ["altAzGridColor", string] |
   ["constellations", boolean] |
   ["constellationBoundryColor", string] |
   ["constellationFigureColor", string] |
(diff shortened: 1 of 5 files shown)


pkgw:
Thanks for the nice summary of the issue here. My main concern is around whether changing some of these types would represent an API break. Looking at `engine/src/index.d.ts`, I believe that some of these relevant settings are already exposed in the `EngineSetting` type, so it seems to me that this change would indeed change the API implicit in that type definition. Also, I think that as-is this PR would need to update that file to remove those declarations.

The approach taken for other similar types is to implement this translation in the messaging layer. For instance, `research-app-messages/src/classic_pywwt.ts` defines the `plotType` setting(s) as a string, and then `research-app/src/settings.ts` has some code to translate between string values and the internal enums. I see that the `color` setting of spreadsheet layers is already handling a `Color` type through this framework.

It's a bit of a pain to add the translations for this stuff, especially since we end up needing to manually mirror the settings enumerations to stringify the necessary types in the currently implementation, but I think that would be a better approach if it seems feasible. The internal engine API should ideally be as strongly typed as possible, I think, while the "lowering" to string representations seems like something to do as we translate between that API and the JSON messaging framework, where those typings aren't available.

Carifio24:
Yeah, changing from `Color` to `string` in the engine internals was definitely my least favorite part of the way this is implemented here. I'd forgotten that the research app had this logic for unpacking messages, but I agree that doing the type translation in the messaging layer is a better way to go. I'll update this to do the color parsing there rather than modifying the engine internals.

Carifio24:
@pkgw I've modified this to use an approach like what you had suggested - that is, converting the string into a `Color` in the research app's message handling layer. The setup here uses the same union type + type info object pattern that's used elsewhere in the repository. I briefly looked into seeing if there was some clever mapped type stuff we could do instead, but wasn't getting anywhere so just went with our tried-and-true method.

PR #259, opened 2023-06-30
Have opacity set annotation batch as dirty

Carifio24:
While working on https://github.com/glue-viz/glue-wwt/pull/96, I noticed some issues with changing the opacity of the line, even though we added support for annotation opacity in #238. Since the opacity would update when other properties were changed, I realized that the issue is that we don't mark the annotation batch as dirty when opacity is updated. This means that opacity updates are only realized after another property that _does_ mark the batch as dirty has been changed. This PR remedies this by marking the batch as dirty when the opacity has changed.

Do we also want to do the same thing for the `Label` and `ShowHoverLabel` properties? It seems like we would, but I don't know enough about what these are used for.

### DIFF
--- a/engine/wwtlib/Annotation.cs
+++ b/engine/wwtlib/Annotation.cs
@@ -97,7 +97,11 @@ namespace wwtlib
         public double Opacity
         {
             get { return opacity; }
-            set { opacity = value; }
+            set
+            {
+                Annotation.BatchDirty = true;
+                opacity = value;
+            }
         }
         string id;
         


pkgw:
Good catch. As for `Label` and `ShowHoverLabel`, I don't know off the top of my head. Setting the dirty flag causes various cached properties of the batch to be recomputed on the next render cycle, so you need to set it whenever any property changes that invalidates those cached properties. Those label settings (presumably!) affect the rendering, but if they don't rely on the precomputed properties, it should be OK not to set the dirty flag.

Carifio24:
I assume by precomputed properties you mean the lists of primitives defined [here](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/Annotation.cs#L26)? If so, it doesn't look like either of the label values affect those - it looks like the `Label` of the relevant annotation (if there is one) is used to determine the hover text [here](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/wwtlib/WWTControl.cs#L115). As for `ShowHoverLabel`, I actually don't see it being used anywhere. So yeah, no need to mark the batch as dirty for either.

pkgw:
Yeah; empirically, it's just "the list of things that are recalculated when BatchDirty is true" :-) So it looks like this is fine as-is.

PR #261, opened 2023-07-10
Drop C#/ScriptSharp and express the engine directly in JS

pkgw:
We are not realistically going to merge the Windows and web codebases again, and there are concerns that the reliance on the extremely-unmaintained ScriptSharp system is going to be constraining. So, let's take the plunge and start building from direct JavaScript inputs rather than the C# files.

Here we are aiming to modify the actual implementation as little as possible, so we import a 45,000-line JS file that represents the output of ScriptSharp. Since they're already in the repo, we keep the C# files around for reference (including comments). The vision is that over time, we'll break the JS into more maintainable pieces and migrate away from some of the awkward constructs that the transpilation model required.


pkgw:
Looks like I need to track down some issues in the test suite, but I posted a draft #262 which builds on this PR and shows how we can start using webpack to bring the JS code back into a more maintainable state.

(After an initial pass of breaking the code back into source files, the nice next thing to do would be to turn it into TypeScript to get all the advantages that that offers.)

PR #262, opened 2023-07-13
ESM+webpackification of the engine code

pkgw:
This pull request builds on #261. It switches the engine module to build with Webpack, allowing us to more conveniently break the code into individual files using ES6 imports and exports (while compiling down to code that's compatible with the ScriptSharp implementation). As a sample I've extracted one of the first modules.

This transition unfortunately involves another large batch of code churn to set things up, but the later commits in this pull request ought to be readable. (Note that, as long as #261 is unmerged, the diffs and commit list for this PR will include the ones in that PR as well.)

Also, we're not importing comments from the astrocalc C# code since it's already awkward code that was auto-translated from C++ in the Microsoft days, and these routines are about as locked-in as code can get.


pkgw:
@Carifio24 @imbasimba OK, I think we have a workable approach here. The combination of ESLint and ESBuild seems to work to help ensure that there aren't any mistakes in the code rearrangements. At the moment I've gone through and rearranged code in the `astrocalc/` subtree, which adds up to about 6500 lines, and then another 5500 lines in the toplevel. I'm feeling good about how this migration process is going.

pkgw:
I have tested as much downstream functionality as I could (web client, pywwt, apps) and I believe this is ready to merge.

PR #265, opened 2023-08-22
Allow multiple instances to be mounted

nmearl:
We've a use case in which multiple instances of the Vue application are being mounted to the same dom (specifically in an ipywidget context with Solara). In this case, however, the id of the div is always the same, causing issues in rendering. This PR allows for passing in a custom id to which the Vue component can be mounted, falling back to the default case of an incremented id when the prop is not provided.

### DIFF
--- a/docs/engine/engine-pinia-index.md
+++ b/docs/engine/engine-pinia-index.md
@@ -91,7 +91,23 @@ createApp(App, {
 
 Note that for now, **you can only include one WWT component in each app**,
 because the WWT engine library maintains global state. To work around this, use
-iframes.
+iframes. 
+
+Alternatively, it is possible to mount separate *full instances* of
+the application to the same web page by passing in a unique `id` when creating
+the Vue app using the `customId` prop.
+
+```ts
+...
+
+createApp(App, {
+    wwtNamespace: "mywwt",
+    customId: "myCustomId"
+  })
+  .use(wwtPinia)
+  .component('WorldWideTelescope', WWTComponent)
+  .mount("#app");
+```
 
 Finally, if you’re using [Webpack], you may run into a pitfall because this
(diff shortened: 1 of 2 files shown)


nmearl:
If you are contributing on behalf of yourself:
@microsoft-github-policy-service agree

If you are contributing on behalf of a company:
@microsoft-github-policy-service agree company="Glue Solutions Inc."

nmearl:
Hey Peter! The use case has to do with interoperability with [solara](solara.dev). In our case, we're wrapping the engine to make something similar to `pywwt`, but since `solara` doesn't use a jupyter kernel, we ended up with a simple widget wrapper around the engine (you can see more [here](https://github.com/nmearl/ipywwt), and my attempt at a workaround [here[(https://github.com/nmearl/ipywwt/tree/unique-instances)).

In this case, every instance of the widget is a Vue app. Therefore, the incrementing does not work, and so multiple instances of the widget attempt to use the same div id. This PR solves this issue by allowing for passing an explicit `id` for the div onto which the Vue app will be mounted.

nmearl:
@microsoft-github-policy-service agree

pkgw:
OK, that makes sense. That's not an ideal setup so hopefully in the long term you won't need to keep using this, but even if this need goes away quickly I can imagine other circumstances where folks could find this useful.

My only other request is that this should be documented. Right now, it looks to me like the best place to do that would be with a mention somewhere in `docs/engine/engine-pinia-index.md`.

The CLA bot seems to be having trouble (https://github.com/microsoft/ContributorLicenseAgreement/issues/143 and it's not detecting your username-@) but it looks like I can merge without needing that to be green, so that'll be OK.

nmearl:
Thanks, @pkgw. I've added some text to the documentation.

PR #274, opened 2023-10-08
Make a pass over the docs, fixing a bunch of problems

pkgw:
Our API docs have gotten pretty busted, between changes to typedoc and the port to Pinia. Here I try to fix them up and improve some wayfinding. Big shout-out to TypeStrong/typedoc#2409 in which our Vue/Pinia component docs were un-broken!


engine-pinia/src/index.ts
```diff
@@ -14,17 +14,45 @@ export {
...
   WWTEnginePiniaState,
-  engineStore
+  engineStore,
 } from "./store";
 
 export { WWTAwareComponent } from "./wwtaware";
 
+// Can't figure out a one-liner way to re-export the default export from
+// Component.vue under a specific name, so:
 import WWTComponent from "./Component.vue";
```
Gerrit0, line 26: `export { default as WWTComponent } from "./Component.vue"`

engine-pinia/src/shims-vue.d.ts
```diff
@@ -1,5 +1,33 @@
 declare module "*.vue" {
   import { defineComponent } from "vue";
+
+  // For reasons unclear to me, the docstring for the WWTComponent component
```
Gerrit0, line 4: This is probably https://github.com/TypeStrong/typedoc/issues/2283, which I haven't gotten to yet...

engine-helpers/src/index.ts
```diff
@@ -353,9 +353,16 @@ class SavedPromise<P, T> {
 }
 
 export class WWTInstance {
+  /** The [`WWTControl`](../../engine/classes/WWTControl-1.html) associated with this instance. */
   readonly ctl: WWTControl;
+
+  /** The [`LayerManager`](../../engine/modules/LayerManager.html) associated with this instance. */
   readonly lm: LayerManagerObject;
+
+  /** The [`ScriptInterface`](../../engine/classes/ScriptInterface.html) associated with this instance. */
```
Gerrit0, line 362: This makes me sad... `@link` *ought* to work here...

pkgw: We have a bunch of separate NPM packages that we manage in a monorepo. From the configuration that we have, I don't see how typedoc would have any way of knowing how to construct a URL to a type defined in (e.g.) `@wwtelescope/engine` while preparing the docs for `@wwtelescope/engine-pinia` — it has no way of knowing that those docs also live in the monorepo and are also built using typedoc.

This basically seems like the problem addressed by Intersphinx in Python-land. If typedoc has any support for solving it, I'd be happy to adopt it, but I didn't see a way to do that built in. It seems like it wouldn't be hard for us to write a plugin to help with linking in our specific case, but I simply haven't sat down to look into that.

engine-pinia/src/index.ts
pkgw: Thanks!

engine-helpers/src/index.ts
Gerrit0: Ah, yeah... that's a problem that TypeDoc *kind of* has a solution for, in that it has the necessary hooks in place that it could be built, but doesn't have a canonical solution -- [third party symbols](https://github.com/TypeStrong/typedoc/blob/master/internal-docs/third-party-symbols.md)

PR #279, opened 2023-10-16
Fix transformations between world and screen coordinates when roll is nonzero

### LINKED ISSUE #273: Screen <---> world transformations behave badly with nonzero roll
This is something that I discovered while working on a CosmicDS mini. The issue is that the screen <----> world coordinate transformations don't give correct results when the camera roll angle is nonzero (as best as I can tell, this affects both directions).

The easiest way to see this is to look at the transformations of the center point of the canvas, which should be invariant under a rotation. One way to try things out is to get a reference to the research app (e.g. `console.log` it in the mounted method and save the output as a global variable), and run the following in your browser console:

```
// Should give the center point of the canvas, as expected
console.log(app.findScreenPointForRADec({ ra: app.wwtRARad * R2D, dec: app.wwtDecRad * R2D }));

const R2D = 180 / Math.PI;
// These two should agree
console.log({ ra: app.wwtRARad * R2D, dec: app.wwtDecRad * R2D });
console.log(app.findRADecForScreenPoint({x: <your-center-x>, y: <your-center-y> }));
```

Now rotate the camera by some amount, without moving the center. Now run
```
// This will no longer give the center RA/Dec
console.log(app.findRADecForScreenPoint({x: <your-center-x>, y: <your-center-y> }));

// This will no longer give the center of the canvas
console.log(app.findScreenPointForRADec({ ra: app.wwtRARad * R2D, dec: app.wwtDecRad * R2D }));
```

As noted in the comments, these will now no longer give the correct values. Even though the roll has changed, the center hasn't, so these results should be unchanged.

I'm not sure yet what the issue is - maybe one of the transformation matrices isn't properly taking the roll angle into account? I'll look into this and see what I can find.

Carifio24:
This PR resolves #273 by making some changes to the transformations between world and screen coordinates. Note that the `Coordinates` functions that are modified here are only used in the `transformPickPointToWorldSpace` and `transformWorldPointToPickSpace` functions.

The reason that this worked before is because when the roll is zero, the view matrix is almost the identity - that fact combined with the minus sign that had been in the coordinate transformations made things work out okay.

This should be easy to test - open up the research app, load up a HiPS catalog, roll the view, and check that the point selection is still working.

### DIFF
--- a/engine/esm/coordinates.js
+++ b/engine/esm/coordinates.js
@@ -199,7 +199,7 @@ Coordinates.cartesianToSpherical2 = function (vector) {
 Coordinates.cartesianToSphericalSky = function (vector) {
     var rho = Math.sqrt(vector.x * vector.x + vector.y * vector.y + vector.z * vector.z);
     var ra = Math.atan2(vector.z, vector.x);
-    var dec = Math.asin(-vector.y / rho);
+    var dec = Math.asin(vector.y / rho);
     return Vector2d.create(ra / Math.PI * 12, dec / Math.PI * 180);
 };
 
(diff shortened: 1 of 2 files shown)


pkgw:
Surprisingly gnarly! Thanks for digging into this.

PR #281, opened 2023-11-03
engine/esm/url_helpers.js: fix HTTPS rewriting corner case

pkgw:
If we were given a URL like `http://mydomain.com:80/` to rewrite in force-HTTPS mode, we'd change it to `https://mydomain.com:80/`, which obviously isn't going to work.

In that one case, it makes sense to rewrite to `https://mydomain.com/` and see what happens. In cases with non-standard port specifications, that's definitely *not* going to work, so in those cases, we just have to keep the URL as insecure HTTP and hope the browser doesn't ding us.

This change motivated by an issue in `ShowImage.aspx`, which seems to have started including the `:80` in the URLs it emits in its redirects, breaking the Astrometry.Net workflow. I'm not sure why that changed, and we could try to change it back, but this modification here makes the rewriting code more correct, and should fully solve the problem. I hope.

@Carifio24 Would you mind giving this a quick read-over and seeing if this seems sensible to you? I want to get this fixed quickly, but don't want to rush a dumb mistake into a release either.

### DIFF
--- a/engine/esm/url_helpers.js
+++ b/engine/esm/url_helpers.js
@@ -237,17 +237,31 @@ var URLHelpers$ = {
             case DomainHandling.tryNoProxy:
             default:
                 if (this._force_https && lcproto !== 'https:') {
-                    // Force HTTPS and we'll see what happens. If
-                    // downloading fails, we'll set a flag and use our
-                    // proxy to launder the security.
+                    // Force HTTPS and we'll see what happens. If downloading
+                    // fails, we'll set a flag and use our proxy to launder the
+                    // security.
+                    //
+                    // As a quasi-hack to work around some stuff emited by the
+                    // Azure backend, make sure to strip off an `:80` port
+                    // specification if we're upgrading. If there's a port
+                    // specification that's anything else, don't try HTTPS - it
+                    // won't work.
                     //
                     // NOTE: it is important that we use `domain` and not
                     // `lcdomain`, even though domain names are
                     // case-insensitive, because we might be processing a
                     // template URL containing text like `{S}`, and WWT's
-                    // replacements *are* case-sensitive. Yes, I did learn
-                    // this the hard way.
-                    return 'https://' + domain + rest;
+                    // replacements *are* case-sensitive. Yes, I did learn this
+                    // the hard way.
+                    var rewrite_domain = domain;
+
+                    if (ss.endsWith(domain, ":80")) {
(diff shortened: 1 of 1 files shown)


Carifio24:
@pkgw Sorry, didn't see this before you merged it in. But the approach here makes sense to me.

pkgw:
Here's an Astrometry.Net/WWT URL to test the use case, pointing at the testing webclient which uses  the latest engine SDK package:

https://worldwidetelescope.org/testing_webclient/?wtml=http%3A%2F%2Fwwtcoreapp-data-app.azurewebsites.net%3A80%2Fwwtweb%2FShowImage.aspx%3Freverseparity%3DFalse%26scale%3D2.815695%26name%3Dtmpdea8lnfx%26imageurl%3Dhttps%3A%2F%2Fnova.astrometry.net%2Fimage%2F20835133%26credits%3DAstrometry.net%2BUser%2B(All%2BRights%2BReserved)%26creditsUrl%3D%26ra%3D324.529527%26dec%3D66.875874%26x%3D2105.0%26y%3D1087.6%26rotation%3D-128.66%26thumb%3Dhttps%3A%2F%2Fnova.astrometry.net%2Fimage%2F20835138%26wtml%3Dtrue

As hoped, this now works. Since I haven't yet made a new release of the engine, the production webclient currently fails:

https://worldwidetelescope.org/webclient/?wtml=http%3A%2F%2Fwwtcoreapp-data-app.azurewebsites.net%3A80%2Fwwtweb%2FShowImage.aspx%3Freverseparity%3DFalse%26scale%3D2.815695%26name%3Dtmpdea8lnfx%26imageurl%3Dhttps%3A%2F%2Fnova.astrometry.net%2Fimage%2F20835133%26credits%3DAstrometry.net%2BUser%2B(All%2BRights%2BReserved)%26creditsUrl%3D%26ra%3D324.529527%26dec%3D66.875874%26x%3D2105.0%26y%3D1087.6%26rotation%3D-128.66%26thumb%3Dhttps%3A%2F%2Fnova.astrometry.net%2Fimage%2F20835138%26wtml%3Dtrue

The problem is that the `ShowImage.aspx` now thinks that its origin is `http://wwtcoreapp-data-app.azurewebsites.net:80`, when it used to get set to `http://worldwidetelescope.org`. This might be caused by the configuration change in the Azure App Gateway frontend to activate Constellations, but the way things are set up I didn't think that would affect this. It's possible that some other configuration change snuck in, or just Azure changed the behavior of something at some point.

PR #283, opened 2023-11-06
Adjust double-touch logic for better usability

Carifio24:
This PR aims to solve some usability issues with the two-touch handling. There are two main changes here:

* If a user has either zoomed or rotated during the current touch sequence, dragging isn't possible until they remove both fingers and start a new touch sequence. This removes the issue that we currently see where it's easy to not pick up both fingers at exactly the same time and get unintended movement after zooming or rotating.
* Allow only one of either zooming or rotating during the same touch sequence. This prevents unintended zooming when rotating, and vice versa. This is a bit more restrictive, so I'm open to other suggestions here. We could always set some sort of timeout (e.g. if you haven't rotated in N milliseconds, unset the rotation flag, and similar for zoom), but I decided to go with keeping things simple.

Since it's also easy to put one finger down and move some epsilon amount before the other one touches, I've also added a  threshold (in pixels) that can be moved before we say that dragging has started to occur.

### DIFF
--- a/engine/esm/wwt_control.js
+++ b/engine/esm/wwt_control.js
@@ -1157,12 +1157,14 @@ var WWTControl$ = {
 
     onTouchMove: function (e) {
         var ev = e;
+
         if (this._hasTwoTouches) {
             var t0 = ev.touches[0];
             var t1 = ev.touches[1];
             var newRect = new Array(2);
             newRect[0] = Vector2d.create(t0.pageX, t0.pageY);
             newRect[1] = Vector2d.create(t1.pageX, t1.pageY);
+
             if (!this._dragging && this._pinchingZoomRect[0] != null && this._pinchingZoomRect[1] != null) {
                 var centerPoint = Vector2d.create(this.renderContext.width / 2, this.renderContext.height / 2);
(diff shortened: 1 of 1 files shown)


pkgw:
This definitely feels a lot better to me. I added one tweak, to adjust the threshold for deciding between rotation and zooming to basically makes it prefer the latter. I think that helps makes things work a bit more naturally.

There's one last piece that I think would help, but I think it can wait for another PR. The current implementation calculates the deltas relative to a `centerPoint` which is just the middle of the viewport, but I think this can lead to some surprising effects. For instance, if I place my fingers at the top-left and bottom-left of the screen and pinch them together, almost nothing happens, even though I'd expect that to lead to a significant zoom-out. Likewise, if my fingers aren't centered around the middle of the viewport, it's very hard to trigger a rotation.

I don't know exactly how other things do this, but I think they're closer to establishing the center point as the midpoint between the two fingers when they first touch the screen. For instance, on Google Maps on my phone, if I place both my fingers in a corner of the screen and rotate, it *does* rotate the view. Notably, though, the axis of the map rotation is the center of he viewport, *not* the center point between my  two fingers.

Anyway, I feel that what we have here is a significant improvement, and I want to make a release to help support my neighbor-arrows work, so I plan to merge this and we can worry about further improvement later.

PR #291, opened 2023-12-01
Allow refreshing tile cache

### LINKED ISSUE #290: Allow clearing tile cache
The Windows client has functionality to allow clearing the tile cache (triggered via pressing F5). It would be useful to have the same sort of cache-clearing functionality in the web engine, although in this case obviously each client could have a different triggering mechanism. This is something we have an occasional need for in CosmicDS, but I think it would be a generally beneficial addition.

I'm happy to look into how to implement this, but any thoughts are welcome.

Carifio24:
This PR resolves #290 by adding functionality to allow refreshing the tile cache. The engine-level implementation is based on the existing functionality from the Windows client [here](https://github.com/WorldWideTelescope/wwt-windows-client/blob/master/WWTExplorer3d/TileCache.cs#L42). I've retained the `try`-`catch` nature of the cache clearing in case some particular tile type's `cleanUp` can throw. This new functionality is also exposed to the Pinia layers and the research app messaging so that it can be used in downstream clients (e.g. `pywwt`).

Since the tile cache isn't really something that clients should be manipulating directly, the only function that I've exposed to the Pinia layer is this new method here. Obviously this doesn't place any restriction at the engine JS level, though.

### DIFF
--- a/docs/engine/research-app-messages-index.md
+++ b/docs/engine/research-app-messages-index.md
@@ -53,3 +53,8 @@ The {@link PingPongMessage} interface is also special. If you send this message 
 app, it will reply with an identical message. This is useful for checking
 whether the app has started up, because usually there is no alternative to
 polling it.
+
+The {@link ClearTileCacheMessage} interface describes a message that, when received by
+the app, will cause it to refresh its current cache of tiles. The primary use case
+for this should be if a network error has caused a necessary tile to fail to load. This
+should be used only when necessary, as it clears all previously downloaded tiles.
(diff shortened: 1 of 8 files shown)


pkgw:
This looks good. My only requests would relate to the documentation: the TS-level and app-message level docs should describe when you might want to use this API and (presumably) give the general message that it should only be used rarely.

The other piece of the documentation is that with these large APIs, we try to maintain more-or-less-formal indexes of all of the methods, so new methods need to be added to those in appropriate places. For the JSON messages, I guess that would be in [the research-app-messages overview][1]. For the Pinia API, that would be in [the engineStore overview][2].

We have enough of an API surface area and enough stakeholders that adding sufficient documentation is really an important part of maintaining the engine codebase.

[1]: https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/docs/engine/research-app-messages-index.md
[2]: https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine-pinia/src/store.ts#L804

PR #292, opened 2024-01-05
keep track of zoomMin

johnarban:
The current version does not check the value of zoomMin when zooming, so nothing changes when it gets set. This just duplicates the logic used for the maximum zoom and uses it for the minimum zoom.


johnarban:
@dotnet-policy-service agree

pkgw:
Thanks John!

PR #304, opened 2024-03-06
Use mode-dependent bounds in gesture handler

### LINKED ISSUE #303: Touch zooming in 3D mode has issues on iOS
Currently the web engine has a problem with touch zooming on iOS (more specifically, in WebKit on a device that recognizes gesture events). In particular, if the camera is out far enough past the Solar System (well before we reach galactic scale), then any zoom motion will bring the camera back close to the Solar System.

The reason for this is that the maximum zoom set for a gesture handler is 360, regardless of the WWT context. 360 is the default max zoom for sky or planetary modes, but is not appropriate at all for 3D mode, where the zoom values get quite large. See the code for gesture handler [here](https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/esm/wwt_control.js#L1134). 

```
onGestureChange: function (e) {
        var g = e;
        this._mouseDown = false;
        this.renderContext.targetCamera.zoom = this.renderContext.viewCamera.zoom = Math.min(360, this._beginZoom * (1 / g.scale));
    },
```

Carifio24:
This PR resolves #303. The change here is to use `get_zoomMax()` to find the context-dependent max zoom (including accounting for user-specified zoom bounds). Additionally, I also added some logic to make sure that the zoom is greater than the relevant minimum zoom as well.

### DIFF
--- a/engine/esm/wwt_control.js
+++ b/engine/esm/wwt_control.js
@@ -1131,11 +1131,10 @@ var WWTControl$ = {
     onGestureChange: function (e) {
         var g = e;
         this._mouseDown = false;
-        this.renderContext.targetCamera.zoom = this.renderContext.viewCamera.zoom = Math.min(360, this._beginZoom * (1 / g.scale));
+        this.renderContext.targetCamera.zoom = this.renderContext.viewCamera.zoom = Math.max(this.get_zoomMin(), Math.min(this.get_zoomMax(), this._beginZoom * (1 / g.scale)));
     },
 
     onGestureEnd: function (e) {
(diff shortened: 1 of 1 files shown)


astrodavid10:
@pkgw can you please push a new release with this fix?

pkgw:
I'll add it to the todo list. How urgent is this?

astrodavid10:
Thanks! Not extremely urgent, but I can't make a functional 3D interactive without manual code futzing or sending to a non-unix machine for a build with @Carifio24's patch.

PR #320, opened 2025-03-12
Allow specifying duration of movement time

Carifio24:
This PR implements something that we want in CosmicDS, which is the ability to specify the duration of a movement between two locations.

As for how I implemented this:
* Add a new optional `duration` parameter to our parameters that control goto movements. This represents the desired duration of the move, in seconds.
* We have two view mover classes: `ViewMoverKenBurnsStyle` and `ViewMoverSlew`. The Ken Burns-style mover already has a parameter for the time to target, so almost all of the work here is for the slew mover. We generally create the slew mover via its `create` function (in fact, I think that's the only way that we do in the codebase), so I've updated that function to take in a `duration` parameter. If the caller doesn't pass anything in, this will be undefined, and the behavior is left unchanged from what it is currently
    - My original plan had been to do something like `createUpDown`, and set appropriate values for the various factors to give the desired duration. However, this isn't possible due to the second term in [this line](https://github.com/Carifio24/wwt-webgl-engine/blob/27bffda195389d0752e8bcf0d6e2c95939607265/engine/esm/view_mover.js#L173), which doesn't have a prefactor that we can adjust. Thus, the approach taken here is a simpler one: create the view mover and let it figure out its times, then adjust the up/down/total times to have the same relative sizes, just scaled to fit the desired duration. This works because only the target time variables are used for the interpolation - the up/down/travel factors are only used in `init`. If not for how they're used in `createUpDown` (whose API I assume we don't want to change), they could just be scoped to that function.
* The rest of the PR is then just plumbing this new functionality through the various layers of the codebase

If anyone wants to test this out, I've set up a demo [here](https://carifio24.github.io/testing/wwt-move-time/) that allows specifying the various parameters. Similar to the time functionality exposed in #222, the actual time for the movement will within a few milliseconds of the requested duration.

### DIFF
--- a/engine-helpers/src/index.ts
+++ b/engine-helpers/src/index.ts
@@ -171,6 +171,9 @@ export interface GotoTargetOptions {
   /** If true, the camera will continue tracking the view target as it moves
    * with the progression of the WWT internal clock. */
   trackObject: boolean;
+
+  /** Optional: The desired duration of the movement */
+  duration?: number;
 }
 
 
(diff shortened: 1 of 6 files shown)


Carifio24:
@pkgw If you're interested in easily trying it out there's a demo (with a UI to set parameters) in the description

PR #330, opened 2025-07-15
Send finder scope place in a message

Carifio24:
This PR follows up on #324 by adding functionality for sending the currently selected finder scope `Place` in a message. As with #324, if the finder scope isn't set to be active this does nothing. Also, like all other research app messaging, this is only sent if there's an appropriate destination to send it to. In the interest of keeping the client up-to-date, we also send a message is the selected place is null (note that this will happen when the finder scope is closed).

For sending the message, the `Place` is serialized as XML (unless it's null, in which case we just send that). To facilitate this, I've added an `asXml` method to the `Place` object which is exposed to TypeScript from the engine package. The implementation of this just hooks into the pre-existing XML serialization for a `Place` object. This allows a receiving client to use existing WWT code for deserialization (whether from the engine, `wwt_data_formats`, etc.) and means that we don't have another serialization format to maintain for a class that has a lot of options.

The intended use case here is pywwt, but our messaging scheme is client-agnostic and this could be used by any client interacting with the research app via the messaging protocol.

### DIFF
--- a/research-app-messages/src/finder_scope.ts
+++ b/research-app-messages/src/finder_scope.ts
@@ -1,5 +1,16 @@
+// Copyright 2021-2023 the .NET Foundation
+// Licensed under the MIT License
+
 // Note: module-level doctring found in `index.ts`
 
+
+/** Information about the current state of the focused `Place` in the research app's
+* Finder Scope.
+*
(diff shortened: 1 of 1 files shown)


Carifio24:
Similar to the message that sends the content of a tour, I've updated the message field here to specifically note that the `Place` as serialized as an XML.

PR #377, opened 2026-04-15
Point depth in 3D

Carifio24:
This PR builds on some hacking that @johnarban and I did for https://github.com/cosmicds/artemis-ii to get points to correctly display in front of/behind planets in 3D mode. The key ideas here are:

* We need to set these layers' point lists to be depth-buffered
* We need to disable writing to the depth mask while drawing point layers. The reason for this is that the Gaussian point blob is actually an image texture with a transparent background. If we write to the depth buffer, things behind the transparent piece of the texture will be clipped when they really shouldn't be. We also don't want to clip points against each other so that we can increase the brightness when points stack. The same is true for point planets.
    - The other option would be to `discard` inside the fragment shader, but this can disable early-Z optimization
* Because of this, we need to draw our layers after we draw the celestial bodies

~~I'm wondering whether the `depthMask` calls should be somewhere in the shader so that we don't need to make sure and track down everywhere that they're used.~~
I moved the `depthMask` calls inside the `draw` and `drawTextured` methods of the `PointList` primitive, so that we don't need to worry about missing it anywhere. This can be adjusted by an optional flag that's added to the function signature.

To see this in action, there's a small example [here](https://carifio24.github.io/wwt-depth-buffer-testing/) demonstrating a set of points in front of and behind the moon.

### DIFF
--- a/engine/esm/baseplanets.js
+++ b/engine/esm/baseplanets.js
@@ -20,7 +20,9 @@ BasePlanets.drawPointPlanet = function (renderContext, location, size, color, zO
         ppList.minSize = 20;
         ppList.addPoint(location.copy(), color._clone(), new Dates(0, 1), size / 100);
         ppList.depthBuffered = true;
+        renderContext.gl.depthMask(false);
         ppList.draw(renderContext, 1, false);
+        renderContext.gl.depthMask(true);
     }
     else {
         var screenSpacePnt = renderContext.WVP.transform(center);
--- a/engine/esm/baseplanets.js
+++ b/engine/esm/baseplanets.js
@@ -20,9 +20,7 @@ BasePlanets.drawPointPlanet = function (renderContext, location, size, color, zO
         ppList.minSize = 20;
         ppList.addPoint(location.copy(), color._clone(), new Dates(0, 1), size / 100);
         ppList.depthBuffered = true;
-        renderContext.gl.depthMask(false);
         ppList.draw(renderContext, 1, false);
-        renderContext.gl.depthMask(true);
     }
     else {
         var screenSpacePnt = renderContext.WVP.transform(center);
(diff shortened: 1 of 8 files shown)


Carifio24:
For a comparison with the current state of things, run the following in the JavaScript console on the webclient:

```javascript
        wwtc = wwtlib.WWTControl.singleton;
        wwtc.gotoRADecZoom(17.75, -28.9, 10, false);

        const iset = "Solar System";
        wwtc.setBackgroundImageByName(iset);
        wwtc.setForegroundImageByName(iset);

        wwtc.renderContext.set_solarSystemTrack(wwtlib.SolarSystemObjects.moon);

        const r = 0.00005;
        const items = [];
        const N = 20;
        for (let i = 0; i < N; i++) {
          const angle = i / 20 * 2 * Math.PI;
          const x = r * Math.cos(angle);
          const y = r * Math.sin(angle);
          items.push(`${x}\t${y}\t0`);
        }
        const data = items.join("\r\n");

        const layer = wwtlib.LayerManager.createSpreadsheetLayer(
          "Sky",
          "Data",
          data,
        );
        layer.set_xAxisColumn(0);
        layer.set_yAxisColumn(1);
        layer.set_zAxisColumn(2);
        layer.set_cartesianScale(wwtlib.AltUnits.astronomicalUnits);
        layer.set_coordinatesType(wwtlib.CoordinatesTypes.rectangular);
        layer.set_markerScale(wwtlib.MarkerScales.screen);
        layer.set_scaleFactor(100);
        layer.set_showFarSide(true);
        layer.set_color(wwtlib.Color.fromArgb(255, 255, 0, 0));
        window.layer = layer;

        wwtc.renderOneFrame();
        wwtc.zoom(0.0003 / wwtc.renderContext.viewCamera.zoom);
```
Currently the points are always blocked by the moon, even when they shouldn't be.

PR #390, opened 2026-05-14
Add fade in/out transition for various display settings

Carifio24:
Inspired by @astrodavid10 doing this for the constellations in a recent interactive, this PR updates the engine to fade certain items in/out. To do this, we check on each frame whether the relevant flag for each of these items has been changed, and if so we bring the opacity up/down by an amount proportional to the number of milliseconds since the transition has started. Once we arrive at the target opacity we unset the start time of the transition. Note that the transition process is handled by the preexisting `BlendState` class.

Along the way I updated the text and simple line shaders to take opacity arguments. If not specified, these are defaulted to 1 for backwards compatibility.

The values that can fade in/out are:
* Constellations (images, figures, boundaries, and names)
* Grids and their text
* Ecliptic and precession chart
* The reticle crosshairs

Below is a video that shows this in action (using a local webclient build)

https://github.com/user-attachments/assets/c7099bcd-6767-42ee-83a0-71f69b1f4d18

### DIFF
--- a/engine/esm/constellations.js
+++ b/engine/esm/constellations.js
@@ -159,7 +159,7 @@ Constellations.initializeConstellationNames = function () {
 // The WWTControl driver will not (and should not) call this function in
 // "freestanding mode", because the functionality depends on a
 // worldwidetelescope.org API.
-Constellations.drawArtwork = function (renderContext, opacity=100) {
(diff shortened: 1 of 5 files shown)


Carifio24:
Rebased to pick up CI changes.

PR #409, opened 2026-06-04
use more robust nan detection

johnarban:
At least for me (running Chrome on Mac with Apple M2), the NaN values in a FITS file are not being detected in the shader. Optimization is basically optimizing the value of the `isNaN` function to be a constant `false` and NaN ends up being collapsed to zero. 

This PR adds a bits based check for nan values. 

I originally used `isnan`. But considering how long it has been a part of the standard, I was concerned about why it wasn't used originally. Apparenlty, if a system does not support NaN it will always return false. Perhaps this is why it was avoided. 

In any case, the NaN is still a value, and by looking at the bits it can be identified.


Carifio24:
Just as a note for posterity - the link in the comment references HLSL (which is for DirectX), but WebGL uses the same IEEE-754 standard so the logic from the post still applies.

This change makes sense to me. Do you have an example of a FITS image with NaN values? I'd like to do a comparison test.

PR #416, opened 2026-06-16
Use same rotation matrix for coordinate transformations and render context

### LINKED ISSUE #415: Camera location after a goto move is slightly off in galactic plane mode
As first noticed by @johnarban: when moving the camera to a specified RA/Dec/Zoom in galactic plane mode, the camera location is slightly off from the specified location. This occurs whether one moves to a location instantly or not.

Carifio24:
This PR resolves #415. The issue is that the `Coordinates` class and the render context have slightly different definitions for the galactic rotation matrix. The approach taken here is to modify the render context matrix to get its entries from the `Coordinates` rotation matrix, so that the coordinate transformations and render context are on the same page. I decided to use the `Coordinates`version as many places might call on the coordinate transformations, whereas the render context matrix is internal to its matrix setup, so this should create the least change overall.


johnarban:
This seems like a good approach. 

As a reference for posterity, the matrix used by `Coordinates` is equivalent to the [Hipparcos-Gaia ](https://gea.esac.esa.int/archive/documentation/GDR1/Data_processing/chap_cu3ast/sec_cu3ast_intro.html#SS7) matrix (just using 10 digits instead of the full 16). It should agree with `astropy` to better than &leq; 0.1 arcsec

Issue #401, opened 2026-05-26
Allow more general filtering of layers


Carifio24:
Something that came up in the course of making an interactive last week was that it would be nice to have a more general way to filter out points from a table layer based on values in the WWT or document state. I'm thinking along the lines of the time filtering, only something a bit more general.

I suspect we would want the implementation to be a bit different, though. The time-based filtering is actually currently done inside the shader, which is fine when we're setting a particular set of columns to be compared against a particular value (the WWT clock), but for general filtration I think we'll need to filter out points in JavaScript when setting up the vertex buffer. Depending on what sort of filter a user specifies, this could potentially slow rendering down.

The other question is what the signature for something like this would look like. My best thought so far is that the filter function could take in a table row, and maybe a reference to the layer and/or the header row as well. Internally the row data is stored as strings, so maybe our filter type could look something like `TableFilter: (row: string[], header: string[], layer: SpreadSheetLayer)`? This would be the main API choice to make here and so deserves some thought.

Carifio24:
Done in #418

PR #418, opened 2026-06-22
Table layer filtering

Carifio24:
This PR is a first attempt at implementing table layer filtering as described in #401. The way that this is done here is as follows. I primarily was trying to minimize storage space and `WebGLBuffer` writes.
* If the user doesn't ever create a filter, no extra space is allocated
* When the user assigns a filter, we create a separate WebGL buffer for the mask (which points are shown or not). The mask booleans are passed in as unsigned bytes. That's the smallest thing that we can pass in as attributes. I guess in the future we could think about some kind of bitpacking, but that's beyond the scope of what I want to do here. We then use this value in the shader - if it's zero, we use a clear color. This approach (using a clear color if out of range) is similar to what's done with the start/end date column calculation.
* When the filter is updated, we update only the mask buffer - the primary buffer for the table data is never touched by any of this filtering scheme. This API here allows the caller to specify whether the filter is static or dynamic. If the filter is static, the mask buffer is updated once - I'm imagining the use case for this is that the filtering is done by a user selecting a UI element or some discrete action. If the filter is dynamic, the mask buffer is updated on each frame. This is a lot more costly, but if the user's filter depends on something like e.g. the viewport position that can change constantly, there's really no other option.

To try out an example of this, add the following inside the `waitForReady` block of the research app:
```javascript
const N = 20;
const items: string[] = [];
for (let i = 0; i < N; i++) {
  const ra = 45 + 2 * Math.random();
  const dec = 20 + 3 * Math.random();
  items.push(`${ra}\t${dec}`);
}

const data = items.join("\r\n");

this.createTableLayer({
  referenceFrame: "Sky",
  name: "Data",
  dataCsv: data,
}).then(layer => {
  layer.set_lngColumn(0);
  layer.set_latColumn(1);
  layer.set_raUnits(RAUnits.degrees);
  layer.set_cartesianScale(AltUnits.astronomicalUnits);
  layer.set_markerScale(MarkerScales.screen);
  layer.set_scaleFactor(100);
  layer.set_showFarSide(true);
  layer.set_color(Color.fromArgb(255, 255, 0, 0));

  function filter(_row: string[], _header: string[], index: number, _layer: unknown) {
    return index % 2 == 0;
  }

  function filterDynamic(_row: string[], _header: string[], index: number, _layer: unknown) {
    return index % 2 == Math.round(Date.now() / 1000) % 2;
  }

  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
  // @ts-ignore
  window.layer = layer; window.filter = filter; window.filterDynamic = filterDynamic;

  this.gotoRADecZoom({ raRad: 45 * D2R, decRad: 20 * D2R, zoomDeg: 50, instant: true });
});
```

Then in the JavaScript console you can run e.g.:
```javascript
layer.set_filter(filter, false);  // The static filter
layer.set_filter(filterDynamic, true); // A dynamic filter that remakes the mask buffer on each frame
layer.remove_filter();  // you could also do `layer.set_filter(null, false)`
```

### DIFF
--- a/engine/esm/graphics/gl_buffers.js
+++ b/engine/esm/graphics/gl_buffers.js
@@ -44,12 +44,26 @@ registerType("IndexBuffer", [IndexBuffer, IndexBuffer$, null, ss.IDisposable]);
 // wwtlib.MaskBuffer
 
 export function MaskBuffer(mask) {
-    this.buffer = tilePrepDevice.createBuffer();
-    tilePrepDevice.bindBuffer(WEBGL.ARRAY_BUFFER, this.buffer);
-    tilePrepDevice.bufferData(WEBGL.ARRAY_BUFFER, mask, WEBGL.STATIC_DRAW);
+    this.mask = mask;
 }
 
 var MaskBuffer$ = {
+    lock: function () {
+        return this.mask;
+    },
+
+    unlock: function () {
+        this.buffer = tilePrepDevice.createBuffer();
+        tilePrepDevice.bindBuffer(WEBGL.ARRAY_BUFFER, this.buffer);
+        var maskArray = new Float32Array(this.mask.length);
+        for (let i = 0; i < this.mask.length; i++) {
+            maskArray[i] = Number(this.mask[i]);
+        }
+        console.log(this.mask);
+        console.log(maskArray);
+        tilePrepDevice.bufferData(WEBGL.ARRAY_BUFFER, maskArray, WEBGL.STATIC_DRAW);
+    },
+    
     dispose: function () {
         tilePrepDevice.bindBuffer(WEBGL.ARRAY_BUFFER, null);
         tilePrepDevice.deleteBuffer(this.buffer);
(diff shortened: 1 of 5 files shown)


Carifio24:
Currently the filter function has a signature like `filter(row: string[], header: string[], index: number, layer: SpreadSheetLayer)`. `row` here is the table row, where internally everything is stored as strings. I also figured having access to the table header and row index might be useful for some filtering applications, as well as the layer instance itself.

Carifio24:
It's also worth noting that the layer filter will **not** be serialized with the layer. This is primarily a technical reason (it's not possible to serialize an arbitrary function, which can in general depend on values from its surrounding scope(s), builtin functions, etc), but I don't think it's really a problem - I'm imagining these filters more as something ephemeral to affect rendering based on the state of the user's application, rather than being intrinsic to the definition of the layer itself.

Carifio24:
Rebased to resolve conflicts

Carifio24:
Rebased again to pick up CI updates

PR #420, opened 2026-06-22
Fix calculation of scaleX from CD matrix for WCS images

### LINKED ISSUE #419: WWT sometimes doesn't go to correct place when loading FITS image
When loading the attached FITS file, the WWT camera will not target the correct location when going to the image after load (e.g. when loading in the webclient). The image appears to render in the correct location, only the goto location is misaligned.

After doing some digging I believe the problem is here: https://github.com/WorldWideTelescope/wwt-webgl-engine/blob/master/engine/esm/layers/wcs_image.js#L228. I think is meant to multiple the sqrt(sum of squares) by the sign of the determinant, but this would need to be wrapped in parentheses - instead, the entire expression is evaluated as the first piece of the ternary, and since that's essentially never going to be zero we'll get -1. And indeed, fixing this does put the camera in the correct location for this image. Digging way back into the history, one can see that this was previously more explicit (I'm guessing the change away from `Math.sign` was to avoid the exceptional case at 0?): https://github.com/WorldWideTelescope/wwt-webgl-engine/commit/ec90eae65cce8cbf9ccf68c5a5a77a725c012045#diff-a2dcdc57f3752f8d0a3f412a78d072e41a57a598d8de9b2eda9649e24e6d93a2

[a01695_mosaic_00_16r90ww.zip](https://github.com/user-attachments/files/29221522/a01695_mosaic_00_16r90ww.zip)

Carifio24:
This PR fixes #419. See that issue for the discussion of the problem.

### DIFF
--- a/engine/esm/layers/wcs_image.js
+++ b/engine/esm/layers/wcs_image.js
@@ -225,7 +225,7 @@ var WcsImage$ = {
     },
 
     calculateScaleFromCD: function () {
-        this.scaleX = (Math.sqrt(this.cd1_1 * this.cd1_1 + this.cd2_1 * this.cd2_1) * (this.cd1_1 * this.cd2_2 - this.cd1_2 * this.cd2_1) < 0) ? -1 : 1;
+        this.scaleX = Math.sqrt(this.cd1_1 * this.cd1_1 + this.cd2_1 * this.cd2_1) * ((this.cd1_1 * this.cd2_2 - this.cd1_2 * this.cd2_1) < 0 ? -1 : 1);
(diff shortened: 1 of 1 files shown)


johnarban:
That is the correct conversion from a CD matrix to the scales (`CDELT1` &  `CDELT2`, in FITS these are x & y pixel scales). A decent reference for the transformations is [wcs 88](https://lweb.cfa.harvard.edu/~jzhao/SMA-FITS-CASA/docs/wcs88.pdf) and references here [FITS Standard](https://fits.gsfc.nasa.gov/fits_wcs.html)

PR #421, opened 2026-06-26
Bump Node setup task in CI; bump browserstack-local version

Carifio24:
This PR bumps `NodeTool@0` to `UseNode@1` in our CI scripts as the former is deprecated - despite the different name, this is the new version of that task (see the Microsoft docs [here](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/use-node-v1?view=azure-pipelines)). We aren't doing anything too fancy so the update is simple, the only real change for our purposes is that the version input is now `version` rather than `versionSpec`.

### DIFF
--- a/ci/azure-job-setup.yml
+++ b/ci/azure-job-setup.yml
@@ -65,10 +65,10 @@ steps:
   displayName: Restore release commit
 
 - ${{ if parameters.setupBuild }}:
-  - task: NodeTool@0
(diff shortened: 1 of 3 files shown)


Carifio24:
There was also an (unrelated) issue with `browserstack-local` that was blocking CI, so I've fixed that here as well.

PR #436, opened 2026-07-31
APIs for adding user-created text

Carifio24:
Something that we've long wanted for CosmicDS, and that I can imagine will be useful to the userbase in general, is an API for generating custom text annotations. Historically this wasn't really possible due to the restricted character set in the spritesheet, so we've monkey-patched in various solutions for the data stories.

With #425 we now have a much more full-featured character set, and as these spritesheets are script-generated they will be easy to update. Thus I think we're at the point where we can offer custom text annotation functionality to users.

This PR is my proposal to do that. It adds functionality for managing text batches and individual text items at both the engine and Pinia/Vue levels. I'd like to think that this is mostly straightforward, but there are a few things worth noting:
* I tried to make as few changes as possible to the engine classes that control the text rendering, so as to keep these changes backwards-compatible with existing interactives
* Similar to, for example, annotations, the engine batches text items together into "text batches" to minimize the number of draw calls required. The draw methods for text batches are what specify the color and size (though individual text items can be scaled relative to this size). This PR preserves that structure and exposes it to the user. As for logical organization, a "text batch" should consist of text annotations that all have similar display properties and, if relevant, should be updated at similar times - an update to any of its constituent text elements will require rebuffering the entire batch.
* Inside the engine I've added a data structure to store user-created text batches. This data structure is keyed by "name" (essentially just an ID for the batch) and also stores the color to use to draw the batch. Note that the alpha channel for the batch draw color is respected.
* As mentioned above, I didn't want to fundamentally change the existing text rendering behavior. With that said, the current sizing within the text framework is (at least to me) a bit awkward. Creating a text batch with height 1 leads to rather large text that spans a good portion of the sky vertically. A lot of the internal engine text height management is a bit strange as well - for example, the text batch for constellation names has height 80, but each constellation name is then scaled by a factor of 0.000125. This would be the equivalent of having a text batch with size 0.01. These labels are roughly 1 degree in height in the sky, I don't think this sizing setup feels particularly reasonable (or comprehensible to users), so the Pinia layer here makes the text sizes be a bit friendlier by dividing the user-input size by 100.


johnarban (approved):
This seems great

PR #438, opened 2026-08-07
Allow drawing annotations in different coordinates

Carifio24:
While the WWT annotation system is very useful, it currently has the restriction that all annotation coordinates must be specified in RA/Dec. This has been a significant pain point for CosmicDS, as we often want to draw things in alt/az coordinates. These coordinates have a time-varying relationship with RA/Dec, which means that we can't just transform once and be done - we've needed to either update the annotation on each frame (requires re-buffering and leads to jittery annotations) or just directly add our own graphics primitives. The latter option works, but requires a rather decent knowledge of how WWT rendering works, and necessitates the objects being drawn at the end of the frame (in a frame callback), or directly modifying the render frame. IMO this points to something missing from the API.

Hence this PR. In order to facilitate a somewhat general approach to using different coordinate systems, I've made a few changes to the annotations setup:
* Currently all annotations in the engine are prepped and buffered together. In order to facilitate the changes in this PR, I added an `AnnotationBatch` class. This behaves similarly to the `Text3dBatch` that the engine uses for text. This also has the advantage that users can now decide which annotations they want to batch together to get re-buffered.
* In order to handle time-varying coordinate frames relative to equatorial (like Alt/Az) without re-buffering, this PR uses the same technique that the engine uses for the alt/az grid, which is to temporarily adjust the render context matrices to be drawing in the relevant frame. This is handled via `worldTransform`, `viewTransform`, and `projectionTransform` at the annotation batch level, which is either a `Matrix3d` or function that maps `RenderContext -> Matrix3d` defining how to adjust the world matrix to do the annotation drawing.
* At the annotation level, we allow defining a static `coordinateTransform` to allow specifying coordinates in other systems like galactic. It's not a problem to define these individually on each annotation as they don't need to be recalculated on each frame render and can thus be computed when prepping the batch.
* As a convenience, we provide two static members of the `AnnotationBatch` class:
  - An `AnnotationBatch.horizontalWorldTransform` function that, when assigned as the world transformation for a batch, will draw the annotations with their coordinates representing altitude and azimuth.
  - An `Annotation.galacticCoordinateTransform` that allows specifying annotation points in galactic coordinates

To preserve backwards compatibility, calling `addAnnotation` without a batch name/value will automatically put the annotation into a default equatorial batch. Similarly, omitting the batch argument in `removeAnnotation` and `clearAnnotations` will lead to the current expected behavior.

I've marked this as a draft as there are still a few things that I'm still working out:
* Currently I have the `coordinateTransform` field for an annotation set as the function that maps the 2D sky coordinates (in the relevant frame) to 3D, as that's what the WWT graphics primitives ultimately want. Alternatively, we could have the coordinate transform map the relevant 2D coordinates to RA/Dec (so trivial in the equatorial frame), and then map the RA/Dec coordinates to 3D internally. The current approach is a bit cleaner behind the scenes and avoids a hardcoded internal preference for RA/Dec, but defining the transformation in sky space might be simpler for users?
* I haven't adjusted annotation hover/click detection yet. This is currently only defined for circles anyway - maybe a more generic implementation there can be partially repurposed for polytope annotations?

### DIFF
--- a/engine/esm/annotation.js
+++ b/engine/esm/annotation.js
@@ -42,6 +42,28 @@ AnnotationBatch.horizontalWorldTransform = function (_renderContext) {
     return mat;
 };
 
+AnnotationBatch.overlayWorldTransform = function (position) {
+  var overlayWorldInitial = Matrix3d.rotationYawPitchRoll(-(position.get_RA() - 12) * Coordinates.RCRA, -position.get_dec() * Coordinates.RC, 0);
+  return function (renderContext) {
+    var world = renderContext.get_world().clone();
+    world.invert();
+    return Matrix3d.multiplyMatrix(overlayWorldInitial, world); 
+  }
+}
+
+AnnotationBatch.overlayViewTransform = function (rotation) {
+  var overlayViewInitial = Matrix3d.lookAtLH(
+    Vector3d.create(0, 0, 0),
+    Vector3d.create(0, 0, -1),
+    Vector3d.create(Math.sin(rotation, Math.cos(rotation), 0)),
+  );
+  return function (renderContext) {
+    var view = renderContext.get_view().clone();
+    view.invert();
+    return Matrix3d.multiplyMatrix(overlayViewInitial, view);
+  }
+}
+
 var AnnotationBatch$ = {
     add: function (annotation) {
         this.items.push(annotation);
--- a/engine/esm/annotation.js
+++ b/engine/esm/annotation.js
@@ -31,6 +31,13 @@ export function AnnotationBatch() {
     this._dirty = true;
 }
 
+// Without this matrix, converting from RA/Dec gives
+// (RA, Dec) -> (270 - Az, Alt)
+// This is a previously known problem - for example, the alt/az grid text
+// explicitly offsets by 6 hours = 90 degrees to account for this
+// This matrix provides the correct transformation.
+// It's done in homogeneous coordinates, but we're just flipping x + z
+// and reversing the sign of both
 AnnotationBatch._horizontalWorldAdjustment = Matrix3d.create(
     0, 0, -1, 0,
     0, 1, 0, 0,
@@ -57,7 +64,7 @@ AnnotationBatch.overlayWorldTransform = function (position) {
     world.invert();
     return Matrix3d.multiplyMatrix(overlayWorldInitial, world); 
   }
-}
+};
 
 AnnotationBatch.overlayViewTransform = function (rotation) {
   var overlayViewInitial = Matrix3d.lookAtLH(
@@ -70,7 +77,7 @@ AnnotationBatch.overlayViewTransform = function (rotation) {
     view.invert();
     return Matrix3d.multiplyMatrix(overlayViewInitial, view);
   }
-}
+};
 
 var AnnotationBatch$ = {
     add: function (annotation) {
--- a/engine/esm/annotation.js
+++ b/engine/esm/annotation.js
@@ -31,6 +31,13 @@ export function AnnotationBatch() {
     this._dirty = true;
 }
 
+AnnotationBatch._horizontalWorldAdjustment = Matrix3d.create(
+    0, 0, -1, 0,
+    0, 1, 0, 0,
+    -1, 0, 0, 0,
+    0, 0, 0, 1,
+);
+
 AnnotationBatch.horizontalWorldTransform = function (_renderContext) {
     var zenithAltAz = new Coordinates(0, 0);
     var zenith = Coordinates.horizonToEquitorial(zenithAltAz, SpaceTimeController.get_location(), SpaceTimeController.get_now());
@@ -39,6 +46,7 @@ AnnotationBatch.horizontalWorldTransform = function (_renderContext) {
     var mat = Matrix3d._rotationY(-raPart);
     mat._multiply(Matrix3d._rotationX(decPart));
     mat.invert();
+    mat = Matrix3d.multiplyMatrix(AnnotationBatch._horizontalWorldAdjustment, mat);
     return mat;
 };
 
@@ -162,10 +170,6 @@ Annotation.defaultCoordinateTransform = function (x, y) {
 }
 
 Annotation.galacticCoordinateTransform = Coordinates.galacticTo3dDouble;
-
-Annotation.horizontalCoordinateTransform = function (alt, az) {
-    return Annotation.defaultCoordinateTransform(270 - alt, az);
-}
 
 Annotation.separation = function (Alpha1, Delta1, Alpha2, Delta2) {
     Delta1 = Delta1 / 180 * Math.PI;
--- a/engine/esm/annotation.js
+++ b/engine/esm/annotation.js
@@ -94,7 +94,7 @@ export function Annotation() {
     this.annotationDirty = true;
     this._opacity = 1;
     this._showHoverLabel = false;
-    this._coordinateTransform = Annotation.defaultCoordinateTransform;
+    this.coordinateTransform = Annotation.defaultCoordinateTransform;
 }
 
 Annotation.defaultCoordinateTransform = function (x, y) {
@@ -190,6 +190,15 @@ var Annotation$ = {
         return value;
     },
 
+    get_coordinateTransform: function() {
+        return this.coordinateTransform;
+    },
+
+    set_coordinateTransform: function(transform) {
+        this.coordinateTransform = transform;
+        this.markDirty(true);
+    },
+
     markDirty: function (dirty) {
         this.annotationDirty = dirty;
     },
@@ -209,7 +218,7 @@ export function Circle() {
     this._fillColor$1 = Colors.get_white();
     this._x$1 = 0;
     this._y$1 = 0;
-    this._coordinateTransform = Coordinates.raDecTo3d;
+    this.coordinateTransform = Coordinates.raDecTo3d;
     Annotation.call(this);
 }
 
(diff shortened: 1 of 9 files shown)


Carifio24:
To see a quick example of this in action, add the following to the top of the `waitForReady` block in the research app with this branch checked out:

```typescript
this.applySetting(["showGrid", true]);
this.applySetting(["showEquatorialGridText", true]);
this.applySetting(["showAltAzGrid", true]);
this.applySetting(["showAltAzGridText", true]);

this.applySetting(["showGalacticGrid", true]);
this.applySetting(["showGalacticGridText", true]);
this.applySetting(["galacticMode", true]);

// this.applySetting(["localHorizonMode", true]);

type This = typeof this;

// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-expect-error Adding things to `window` for convenience
window.app = this; window.wwt = WWTControl.singleton; window.addAnnotations = function (this: This) {

    // The horizon
    const n = 20;
    const delta = 360 / n;
    const horizontalBatch = this.createHorizontalAnnotationBatch("horizontal");
    for (let i = 0; i < n; i++) {
      const triangle = new Poly();
      triangle.set_fill(true);
      triangle.addPoint((i + 1) * delta, 0);
      triangle.addPoint(-i * delta, -90);
      triangle.addPoint(i * delta, 0);
      triangle.set_lineColor("#2E6F40");
      triangle.set_fillColor("#2E6F40");
      triangle.set_opacity(0.7);
      this.addAnnotation(triangle, horizontalBatch);
    }
  
    // Galactic coordinates "rectangle"
    this.createAnnotationBatch({
      name: "galactic",
    });
    const galacticRect = new Poly();
    galacticRect.set_fill(true);
    galacticRect.set_opacity(0.8);
    galacticRect.set_fillColor("pink");
    galacticRect.addPoint(0, 0);
    galacticRect.addPoint(0, 10);
    galacticRect.addPoint(10, 10);
    galacticRect.addPoint(10, 0);
  
    galacticRect.set_coordinateTransform(Annotation.galacticTo3dTransform);
    this.addAnnotation(galacticRect, "galactic");
  
    // Static "overlay" annotation that follows the viewport
    const overlayBatch = this.createOverlayAnnotationBatch({
      name: "overlay",
      position: { raRad: 0, decRad: 0, rollRad: 0 },
      rollWithCamera: true,
    });
  
    const staticRect = new Poly();
    const halfSize = 2;
    staticRect.set_fill(true);
    staticRect.set_opacity(0.8);
    staticRect.set_fillColor("maroon");
    staticRect.addPoint(-halfSize, -halfSize);
    staticRect.addPoint(-halfSize, halfSize);
    staticRect.addPoint(halfSize, halfSize);
    staticRect.addPoint(halfSize, -halfSize);
  
    this.addAnnotation(staticRect, overlayBatch);

    const staticRect2 = new Poly();
    const halfSize2 = 7;
    staticRect2.set_fill(true);
    staticRect2.set_opacity(0.8);
    staticRect2.set_fillColor("orange");
    staticRect2.addPoint(-halfSize2, -halfSize2);
    staticRect2.addPoint(-halfSize2, halfSize2);
    staticRect2.addPoint(halfSize2, halfSize2);
    staticRect2.addPoint(halfSize2, -halfSize2);
  
    this.addAnnotation(staticRect2, overlayBatch);

    // this.setClockRate(10_000);

    const circle = new Circle();
    circle.setCenter(0, 0);
    circle.set_radius(5);
    this.addAnnotation(circle);

}.bind(this);

 
```

One can then do whatever settings manipulation in the console (using `window.app`), and then add the annotations with `addAnnotations()`. This should show a green annotation representing the horizon at 70% opacity, and a pink annotation that fills a "square" in galactic coordinates at 80% opacity (I made these partially transparent for easy comparison with the grid coordinates).

Carifio24:
A quick sample of these annotations in action:

https://github.com/user-attachments/assets/9eb5d336-f814-4f7f-b77a-5d6a1bef0b6f

engine/src/index.d.ts
```diff
@@ -512,10 +516,31 @@ export class Annotation implements AnnotationSettingsInterface {
...
+  get_coordinateTransform(): AnnotationCoordinateTransform;
+  set_coordinateTransform(transform: AnnotationCoordinateTransform): void;
 
   hitTest(renderContext: RenderContext, ra: number, dec: number, x: number, y: number): boolean;
 }
 
+export class AnnotationBatch {
+    static readonly horizontalWorldTransform: BatchTransform;
+    static readonly overlayWorldTransform: BatchTransform;
+    static readonly overlayViewTransform: BatchTransform;
```
johnarban: Need to be cast as factory functions. Somewhere, perhaps in a comment, it would be good to note that the usual starting place is "Coordinates.fromRaDec(0,0)". I'm not quite sure what it actually defines (the location of the polygon's (0,0) coord in the "overview static celestial frame?")
```suggestion
    static overlayWorldTransform(position: Coordinates): BatchTransform;
    static overlayViewTransform(rotation: number): BatchTransform;
```

engine/esm/annotation.js
```diff
@@ -9,11 +9,157 @@
...
+    world.invert();
+    return Matrix3d.multiplyMatrix(overlayWorldInitial, world); 
+  }
+};
+
+AnnotationBatch.overlayViewTransform = function (rotation) {
+  var overlayViewInitial = Matrix3d.lookAtLH(
+    Vector3d.create(0, 0, 0),
+    Vector3d.create(0, 0, -1),
+    Vector3d.create(Math.sin(rotation, Math.cos(rotation), 0)),
```
johnarban: missing paranthesis caused matrix to be mostly NaN's
```suggestion
    Vector3d.create(Math.sin(rotation), Math.cos(rotation), 0),
```

engine/esm/annotation.js
```diff
@@ -9,11 +9,157 @@
...
+    var decPart = -(zenith.get_dec() / 360 * (Math.PI * 2));
+    var mat = Matrix3d._rotationY(-raPart);
+    mat._multiply(Matrix3d._rotationX(decPart));
+    mat.invert();
+    mat = Matrix3d.multiplyMatrix(AnnotationBatch._horizontalWorldAdjustment, mat);
+    return mat;
+};
+
+AnnotationBatch.overlayWorldTransform = function (position) {
+  var overlayWorldInitial = Matrix3d.rotationYawPitchRoll(-(position.get_RA() - 12) * Coordinates.RCRA, -position.get_dec() * Coordinates.RC, 0);
```
johnarban:
```suggestion
  var overlayWorldInitial = Matrix3d.rotationYawPitchRoll(-(position.get_RA() - 6) * Coordinates.RCRA, -position.get_dec() * Coordinates.RC, 0);
```

johnarban (changes requested):
This is really cool. I think the api's and interface are ok. But I found a few issues as I tried to play around with this. Some are in the code suggestions, but the main bug  was related to how `AnnotationBatch` handles dirty `Annotations`

If found that I could not change an annotations color opacity, without manually marking the batch as dirty. And then, if a batch had more than 1 annotation, then the annotation that was not changed (not dirty) was not drawn. 

This comes down to this check in the `draw` functions in `annotation.js`

```js
draw: function (renderContext, batch) {
  if (renderContext.gl != null) {
      if (this.annotationDirty) {
        batch.markDirty(true);
```

these should be like they were on the old version, and pass if anything in the batch is dirty (since `prepareBatch` clears all the geometries first, not dirty Annotations do not get drawn). I suggest

1) change the check in the `draw` calls to 
```js
draw: function (renderContext, batch) {
  if (renderContext.gl != null) {
      if (this.annotationDirty || batch._dirty) {
          // batch.markDirty(true);
```

2) remove `this.markDirty(false);` from `prepareBatch`. It should probably only be unmarked as dirty in `drawBatch` anyway

these allow you to do something like
```js
let ann: Annotation
let batch: AnnotationBatch
waitForReady().then(() => {
  batch = this.createAnnotationBatch(...)
  const pol = new Poly()
  ...
  this.addAnnotation(pol, ...)
  ann = pol
})

// somewhere else
const updateAnn() => {
  ann.set_opacity(0.5)
  ann.set_color("#ff0000")
  batch.markDirty(true)
}
```


Preferrably, however, I think we shouldn't need to mark the batch dirty ourselves. I think the best thing would be for each `Annotation` to know what batch they are in. The `Annotation` can have a `_batch` property that can be set during `AnnotationBatch.add`, and then the `Annotation.makeDirty` can call it's  `_batch.makeDirty(true)`. Alternatively, `prepareBatch` could check if it is dirty and if not check for any dirty annotations. 


A couple enhancements that might be nice (if not difficult)
- and ecliptic world tranform built in (since it is also time-varying)
- as nice dev experience thing. my editor picks up /** blah */ comments before functions as docs to show on hover. otherwise just having a code comment in on the code about when and how coordinate transforms vs the world/view transform are used (which gets called on the annotation vs in the `transform` )

engine/src/index.d.ts
Carifio24: Fixed in https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/438/commits/b4d90fab81f4318db4f2890e29d7dd67e75d2616

engine/esm/annotation.js
Carifio24: Fixed in https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/438/commits/48e5501013aebb6e8b9421a0eb6c243c056cac23

engine/esm/annotation.js
Carifio24: Fixed in https://github.com/WorldWideTelescope/wwt-webgl-engine/pull/438/commits/48e5501013aebb6e8b9421a0eb6c243c056cac23

Carifio24:
I've fixed the typos/syntax issues that you noticed, thanks!

I've updated the annotation batch drawing/handling so that clean batch items will still get drawn when another annotation in the batch is dirty. I'm typically not a huge fan of child elements having references to their container unless it's really necessary, so instead I have the batch check whether any of its children are empty when it's preparing the batch. The O(N) scan is not ideal, but unless someone has a LOT of annotations in one batch I can't see this being any sort of bottleneck. I tried this out with a multi-annotation batch in the research app and it worked well for me.

I like the idea of adding in ecliptic coordinate transformations and some helpful docstrings, I can add both of those in before we merge this.

johnarban (approved):
I made one change in 1c75dddd49def9dfaceb1a9275cc3ba0ced87339. Otherwise, this looks good to me

PR #457, opened 2026-09-14
Add new circle markers for drawing points

Carifio24:
This PR adds a new "filled circle" plot type using a new shader for drawing point circles. This approach was first used (in a very customized form) by @astrodavid10 in https://github.com/cosmicds/exoplanet-sonification, and made more generally WWT-compatible in https://github.com/cosmicds/almagal.

This is done via the addition of a new shader for drawing the circle. Unlike the texture-based shaders that we currently have for point drawing, this uses an approach that combines `gl_PointSize` and `gl_PointCoord` to draw the circle in the fragment shader based on the distance of the pixel from the clip space center. When WebGL2 or the [`OES_standard_derivatives` extension](https://developer.mozilla.org/en-US/docs/Web/API/OES_standard_derivatives) are available, we use `fwidth` to dynamically determine the border width; for WebGL1 systems without this extension we provide a fixed-size fallback.

This shader allows drawing the circle with or without a border; I've added a new `bordered` flag to spreadsheet layers and point lists for this, with the caveat that this option currently has no effect for other plot modes.


johnarban (commented):
This looks great. 

The research app "marker" selector in `SpreadsheetItem.vue` doesn't have filled circle in it (that list gets custom built) and i noticed that in `time_series_layer.js` `PlotTypes` is defined without `filledCircle`. 

it does do a weird thing if I change the color on a spreadsheet layer in the research app, the border turns into more of a "squircle"

Image

Carifio24:
Good catches! These should all be fixed with the latest two commits.

